PR #3, opened 2015-07-24
Added get_symmetry_from_database() to Python module.

CarstenTUBT:
Would love to have this function. As I don't use ASE, I didn't write a native python function for that - sorry.

Also, I don't know any C at all. While this works for me, you should review it carefully.


atztogo:
Thanks. Merged to the master branch. I changed the order of variables so that the return values come in front and added checks of the array sizes. I updated ASE example to test it. To see how it works, you can run example.py, for which you don't need to use ASE but can use local code of atoms.py.

PR #4, opened 2015-12-16
Conda and pypi build setup.

jochym:
This diff adds build info for pypi and conda package repositories. I have also fixed relative import for python 3. The package does not build on py3 any more. Problem with symbols in c-extension. It was building fine on 1.8.2.2.


atztogo:
Merged.

PR #8, opened 2016-03-06
Travis CI and deploynment config

jochym:
This is a working code for continous integration/testing/deploynment to pypi. I have tried to minimize changes to the upstream code - the only important change is adding nanoversion (fourth digit) to the version number - this is a build number - we need it for automatic deploynment to pypi - they do not accept any re-uploads with the same version number. I hope I did not broke anything ;)

The deploynment key will probably not work from a different travis account. This is not a pasword.
We can probably transfer the deploynment to your accounts in the next iteration. Please keep it there for now - it will save me time during merges and costs you nothing.


blokhin:
Awesome! 1.9.1 is now ready to be released?

atztogo:
Thanks. I merged to the 'develop' branch. Though I have not understood this is well yet, I checked setup.py went well without any problem.

atztogo:
I will probably soon make a release branch for v1.9.1 after adding/updating some python-spglib APIs.

jochym:
Good. The idea is quite simple. Each push to the repo triggers build on travis and if it is a master branch (release/hotfix) the result gets deployed to the pypi with new number (under certain conditions - one is passing all tests). The versioning is a bit convoluted due to the fact that we need to number post-release versions but cannot use local versioning (+patch number). So I adopted last number as a patch number. So the package is (whatever Togo uses as a release number).patch_number (essentially this is build number). On conda the same scheme uses +patch_number at the end. 
The missing part is testing. I have used the python/test code for testing but there is no testing for c library. Please Togo: review build/testing code in travis.yml and tell me if I can improve it in any way (e.g. testing for C library).

atztogo:
For the test in C, I updated the autotools setup, configure.ac and Makefile.am. By running "make check", src/test.c is compiled and executed, and I think it returns 0 if succeeded and 1 if failed.
I updated .travis.yml for this update. However I have no idea if it is done correctly or not at all. Could you check it?
Both for C and Python, spglib tests were not well written. I will write the tests little by little from now.

blokhin:
Please note, that as of now both commands:

```pip install spglib==1.9.1```
```pip install spglib==1.9.0.54```

do not work. Though a workaround is possible, e.g.:

```pip install -e "git+https://github.com/atztogo/spglib.git@91e8a317ab7e2da8e0d8546b31f40277e697b153#egg=spglib&subdirectory=python"```

which however requires newer version of pip (>7.1).

atztogo:
Thanks for your information. Deployment to PyPI is under testing. Please wait for a while.

PR #11, opened 2016-05-11
Working CI/deploynment setup for 1.9.3

jochym:
These are my modifications necessary to make the automatic release through TravisCI work. 
A lot of the complication comes from the fact that it is working from my tree instead of yours ;)
If you will adopt these changes - you will have to make modifications and some of the complications could be removed completely. I am making this pull request as an easy way to show you the changes. You should probably just review the diffs and close the request (or just leave it open for future work). I can make appropriate simplifications but have no way of testing them in my tree so we will need to cooperate closely on this. For now let us check if this will work for 1.9.4. Alternatively we can try to move it to your tree completely for 1.9.3+ and test if it works on 1.9.4. The choice is yours.


atztogo:
I have two quesitons.
1. When running setup.py manually, ln -s ../src is not done, so probably it fails. Is this symbolic link necessary for travisCI? If it is, then I will insert some lines to make some trick (checking availability of src directory etc).
2. __nanoversion__.txt is under the git control. This is dynamically generated by travisCI, but I guess it's not updated in the git repository by travisCI (this I don't know...)  This file is in MANIFEST.in, so this should exist. Considering these, Is it OK to include to the git repository just as a blank file?

jochym:
Ad. 1: The problem is that the distribution made by setuptools for pypi contains only the python subdirectory so the setuptools makes the copy of the src in python. This is fine if we reference src/*.{c,h} in setup.py but not if we use ../src/ (there is no ../src when you do pip install - and the user install was failing exactly for this reason). I am not very good in hacking setuptools - if you have other ideas how we can make it work both for github clone/zip install and for pip install go ahead. That what I could construct. Note however that the user installing from zip/clone will be probably more technical then the one using pip or conda. So we can just stick 'make a link' instruction in readme and it will be ok. But we definitely need to make pip and conda packages working flawlessly - otherwise what is the point of packaging?

Ad. 2: It is indeed is. I wanted for it to be there to be able to run setup.py without runing get-nanoversion script. It shoud be a file with zero in it (actual number 0). The problem is that pypi does not allow you to upload the same version *ever*. So we need some constantly incrementing build number. I came up with the number of commits after last release tag (git describe --tags)- this way it is at least meaningful. It turns out not to be that easy to use. Maybe we should switch to setuptools_scm instead of cooking our own solutions to known problems?

jochym:
I have added the same tricks as used on phonopy to the pull request. I have tried to use `graft ../src` in MANIFEST.in but it did not work. The rest is just improvements in build pipeline.

atztogo:
I merged your commit. I modified two file, __nanoversion__.txt to contain 0, and setup.py. In setup.py, os.path.exists("src") is checked. I hope this doens't make trouble in travisCI.

PR #12, opened 2016-05-13
Small post-release correction to get_nanoversion

jochym:
It only impacts travis/conda packaging scripts.


atztogo:
Thanks. I merged to master and develop branches. (Was it OK?.)

jochym:
Yes. Great. We are finally in sync with packaging - it should be easier from now on.
Your releases merge cleanly and build without problems.

PR #18, opened 2016-08-29
Add requirements.txt

shyuep:


atztogo:
In requirements.txt, specific version is necessary?

shyuep:
Yes. It is usually recommended that requirements.txt pin the version specifically. This is different from setup.py, which does not pin versions. See http://nvie.com/posts/pin-your-packages/

atztogo:
Merged to develop branch.

PR #21, opened 2016-09-17
Check that config_var is not None. Otherwise setup.py fails on some platforms

shyuep:


atztogo:
Thanks. I merged it to the develop branch.

PR #24, opened 2016-10-23
Modify setup.py to dynamically install numpy before building spglib.

shyuep:
These changes allow spglib to be installed on a system which does not have numpy installed. It uses setup_requires. However, we need to modify the extension build to modify the numpy include (this is a well-known issue) to enable the build.


atztogo:
I don't understand what you mean very much, but probably it is necessary for you. So I just included your change to master branch with some modifications. I preserved the chance to use distutils if setuptools is not installed. I wish my modifications doesn't break your case.

The code was deployed to PyPI and conda. At least this minor update worked for me, but I don't know for you.

shyuep:
I add indent and also added a missing import sys

PR #29, opened 2016-11-03
Exclude test/*.pyc files from package

jochym:


atztogo:
Thanks. I git-rebased master in develop and rc branches.

PR #31, opened 2016-12-07
Update python-spglib.rst to account for hall_number

afonari:
This documents 'hall_number' key that is returned from get_symmetry_dataset


atztogo:
Thanks. With a few more words added, I updated the web site.

PR #44, opened 2017-12-22
optimizing is_overlap_all_atoms: take 1

ExpHP:
Okay.  This patch is in a bit of a confused state right now, but I will put it up for review because I'm not sure which direction to take it right now:

**General thoughts/notes:**
* This is orders of magnitude faster on large structures, but very likely slower for small structures.
* ~~In my profiling, **now almost all time is spent sorting.**~~
  * this is fixed with the second commit
* `PermFinder` stores information about the sorted original cell (which is a good idea), as well as preallocated memory for temporary data. (which was a bad idea)
  * (note: my only reason for preallocating memory was to make error handling easier; I doubt it impacts performance here)

What is in this currently:
* **Present:** [Sorting of positions](https://github.com/ExpHP/spglib/blob/overlap-optimize/src/permutation.c#L111)  by lattice distance.
* **Present:** [Saving data](https://github.com/ExpHP/spglib/blob/overlap-optimize/src/symmetry.c#L552-L556) about the original cell in `get_translation_part`, so that it can be reused across many overlap checks (by calling `perm_finder_check_total_overlap` instead of `is_overlap_all_atoms`).
  * I tried doing this in other places too, but `get_translation_part` is the only place where it made a difference
* **Not present:** Same optimization, but for the [`_OPENMP` branch](https://github.com/ExpHP/spglib/blob/overlap-optimize/src/symmetry.c#L483) in `get_translation`.
  * I did not do this because the preallocated work memory in `PermFinder` makes it not threadsafe.  (Like I said; that preallocated memory was probably a bad idea)
* ~**Present:** Code for computing the [full permutation](https://github.com/ExpHP/spglib/blob/overlap-optimize/src/permutation.c#L411-L429) of an operator (`perm_finder_find_perm`), because it was tricky to get this correct.~
  * (this was present in 920d13a5ea but has been removed)
* **Not present:** Optimizations to `search_pure_translations` to use the perm, because it turns out they weren't necessary.  (as a result, no code currently uses the permutation computed by `perm_finder_find_perm`...)

So right now, the patch feels kind of incomplete and looks unnecessarily complicated, but I'm not sure where to go with it.  Let me know what you think.


atztogo:
Thanks. I will read your patch. I'm sorry that it may take some time to review. I am a bit busy for the other job. I think OPENMP part of the code is usually not used, so it can be removed to make the code simpler.

ExpHP:
I pushed one small additional change that checks a few atoms by brute force before performing the sort.  This reduces the overall cost of spglib by another factor of 10 on my test case, making it fast enough that I can no longer measure it over the rest of phonopy.

(this optimization unfortuntately will not help much for the common case of having a large supercell with a single defect.  But I don't think there is much that *can* be done to help that case in general...)

atztogo:
Thanks. I will see your code. On my mac, I ran tests `python -m unittest discover -b` on `python/test` directory, some of them failed. So something must be wrong. But probably it is related to crystallography issue and I may be able to fix it.

ExpHP:
Hm, that is strange. :/  What I see is:

```
$ python -m unittest discover -b                                                                                                                             
..................                                                                                                                                           
----------------------------------------------------------------------                                                                                       
Ran 18 tests in 46.574s                                                                                                                                      
                                                                                                                                                             
OK
```

atztogo:
I also tried to run tests on ubuntu 14.04 on python2.7 and 3.6 conda environments, but the tests failed. I put a line 'permutation.c' in setup.py. Do I need to do more to correctly compile it?

ExpHP:
Oops, it did not occur to me that I needed to install it for the tests.

I now see 8 or so errors that all suggest that the dataset returned from `spglib` is `None`.  Hm...

ExpHP:
I found the bug and fixed it.

`find_perm_near_identity` was mistakenly allowing atoms of different types to overlap.

atztogo:
Thanks. It worked after you fix. Even for the test cases of the small unit cells, it became faster ~10%. It is nice.

atztogo:
I'm sorry to be late to respond this pull request. I have roughly read your code. It is complicated but I think I can be going to merge it.

spglib is currently written to support C89. ``inline`` keyword is not compiled in some environment (e.g. windows binary of conda-forge package).

Some functions are written as global but only used inside ``permutation.c``. Is there any reason, e.g. performance issue?

Is ``permutation`` an appropriate name?

I don't yet understand well how to measure distance for sorting when rotation is applied. But I guess distance to use for sorting is not necessarily to be measured in ``lattice`` metric. Simply fractional coordinate distance may be used.

I removed OPENMP part in ``symmetry.c``.

I pushed my minor changes to https://github.com/atztogo/spglib/tree/ExpHP-overlap-optimize.

ExpHP:
> spglib is currently written to support C89. `inline` keyword is not compiled in some environment (e.g. windows binary of conda-forge package).

Hm, that's a shame.  Actually, a number of the functions in `mathfunc` would seem to benefit significantly from inlining (since it allows the compiler to reason better about aliasing and alignment), but simply making them `static` without `inline` results in spurious unused warnings.

> Some functions are written as global but only used inside permutation.c. Is there any reason, e.g. performance issue?

> Is `permutation` an appropriate name?

These details are kind of leftovers from an earlier design.  `permutation.h` was originally like this:

```c
typedef void * Argsorter;
Argsorter perm_argsort_work_malloc(int);
void perm_argsort_work_free(Argsorter);
int perm_argsort(int *, const int *, const double *, Argsorter, int n);

void perm_permute(void *, const void *, const int *, int, int);
void perm_compose(int *, const int *, const int *, int);
void perm_inverse(int *, const int *, int);

static inline void perm_permute_int(...) { ... }
static inline void perm_permute_double(...) { ... }
static inline void perm_permute_double_3(...) { ... }
```

and all the code that used it was in other files.  Then at some point I decided to introduce the PermFinder type, and that's when all the code that uses it either got moved into `permutation.c` or deleted.

So yeah, they could be made private.  I only left it this way because I still like the idea of permutations as a reusable abstraction (it *could* be used by other code, it just isn't currently).

> I don't yet understand well how to measure distance for sorting when rotation is applied. But I guess distance to use for sorting is not necessarily to be measured in `lattice` metric. Simply fractional coordinate distance may be used.

I use the cartesian distance to the nearest lattice point. (or I guess [square distance](https://github.com/atztogo/spglib/pull/44/files#diff-2820e41c4bcb5a1953d29daeffffafb6R325))  I would never (purposefully) try to compute distance in fractional space, because that's just too strange. (even for something like sorting, I probably wouldn't trust it to have the right mathematical properties)

ExpHP:
* I incorporated your changes
* I tried to simplify things by removing some unused parts: 5c9c4c2de93f58dcc076b0af221fbef133ce5952...fd2a87bc3f4807b1e41a5139e847a6fc47ea2899
* I renamed the files to `overlap.c` and `overlap.h` and made a lot of stuff local

atztogo:
Thanks. It is fun to read your code.

I merged it to develop branch to go to merge it to master. Can I append a prefix, e.g., ovl_, to the global functions?  (maybe strange custom...)

> but simply making them static without inline results in spurious "unused function" warnings.

If it is true for many compilers, it is fine. Or maybe we can introduce macro to switch them. I have no experience to use windows visual studio C compiler. I heard that it behaves strangely sometimes. spglib is used by many simulation packages. This means that it has to be compiled on different compilers on different supercomputers. For those, a legacy way of writing is still safe for me.

Space group type is an object of group theory, where Cartesian distance is not used. So it's not very strange for me to use fractional coordinates. I think current way of your implementation to measure distance may not consider periodicity of lattice very well. For oblique lattice, it is not aligned from shortest distance. But if I understand correctly, the distance is used just to distinguish atoms, so if they are well labeled, employing length of distance in any coordinates works, since I know the fact that the rotation matrices are chosen after determining lattice point group (i.e., among rotation matrices satisfy the lattice symmetry.)

ExpHP:
On further thought I think everything you're saying is correct.

In previous discussions I said that the reason I chose distance from the nearest lattice point is because it is invariant (except numerically) under rotations; and I also figured it would not work well for translations.  However, this is not actually true.

In reality, the properties that *are* important are:

* It should be a continuous, periodic function
* There should be very few "ties" for the same value across all sites.

ExpHP:
> Can I append a prefix, e.g., `ovl_`, to the global functions? (maybe strange custom...)

Please feel free. I agree this is important to do for c code. (and ovl sounds fine to me)

ExpHP:
I just remembered something.

> I think current way of your implementation to measure distance may not consider periodicity of lattice very well.

I believe I am computing this distance in the same manner that spglib was already doing to compute nearest image distance in its overlap-checking routines.  If this is not valid then it seems to me that this is a pre-existing correctness issue in spglib. (or am I mistaken?)

Maybe a comment should be added to `overlap_checker_init` saying that it expects the cell to have a reduced lattice matrix. (because you are right that this function is **not correct** for arbitrarily skewed lattice matrices)

src/overlap.c
```diff
@@ -0,0 +1,529 @@
...
+    type_rot = checker->types_sorted[i_test];
+    mat_multiply_matrix_vector_id3(pos_rot, rot, checker->pos_sorted[i_test]);
+    for (k = 0; k < 3; k++) {
+      pos_rot[k] += test_trans[k];
+    }
+
+    /* Brute-force search for the rotated position. */
+    /* (this could be optimized by saving the sorted ValueWithIndex data */
+    /*  for the original Cell and using it to binary search for lower and */
+    /*  upper bounds on 'i'. For now though, brute force is good enough) */
```
ExpHP: **Note:** The "this could be optimized" comment should be revised or removed if we switch to computing fractional-space distances because it assumes we are using cartesian distance.

Basically, it relies on the fact that, if two points are within a distance of SYMPREC to each other, then their values for the (non-squared) cartesian lattice point distance will also be within SYMPREC of each other.  This means that if the lattice point distance of `pos_rot` is `rot_value`, we only need to look at the region of values `rot_value - SYMPREC * FUZZ <= value <= rot_value + SYMPREC * FUZZ` (where `FUZZ` is something like `1 + 1e-7`)

(note: to prove the above property about cartesian lattice point distance, notice that the gradient always has a magnitude of 1 except at discontinuities, and that the value is still continuous even where the gradient is not)

atztogo:
> I believe I am computing this distance in the same manner that spglib was already doing to compute nearest image distance in its overlap-checking routines. If this is not valid then it seems to me that this is a pre-existing correctness issue in spglib. (or am I mistaken?)

My explanation was not enough and was misleading. I only thought about measuring distance between atoms for sorting. For overlap checking,  Euclidean distance is necessary because it is a good measure for our intuition.

In phonopy, we need to have Wigner-Seitz cell of supercell. We normally have to consider up to several nearest neighbor lattice points to find the shortest one among supercell-translationally equivalent atoms, for which Euclidean distance was necessary. In spglib, there was no reason to do it.

atztogo:
I merged it to the master branch and it is released as v1.10.3.

atztogo:
Jochym found ``overlap.c`` is not able to be compiled using Microsoft Visual Studio 9.0, which is needed to create windows binary for conda-forge distribution of spglib (and so phonopy).
https://ci.appveyor.com/project/conda-forge/spglib-feedstock
```
C:\Program Files (x86)\Microsoft Visual Studio 9.0\VC\BIN\cl.exe /c /nologo /Ox /MD /W3 /GS- /DNDEBUG -Isrc -IC:\bld\spglib_1516043744425\_b_env\include -IC:\bld\spglib_1516043744425\_b_env\PC -IC:\bld\spglib_1516043744425\_b_env\lib\site-packages\numpy\core\include /Tcsrc/overlap.c /Fobuild\temp.win32-2.7\Release\src/overlap.obj
overlap.c
src/overlap.c(176) : warning C4028: formal parameter 4 different from declaration
src/overlap.c(176) : warning C4028: formal parameter 5 different from declaration
src/overlap.c(331) : error C2036: 'const void *' : unknown size
src/overlap.c(332) : error C2036: 'void *' : unknown size
src/overlap.c(395) : error C2036: 'void *' : unknown size
src/overlap.c(396) : error C2036: 'void *' : unknown size
src/overlap.c(397) : error C2036: 'void *' : unknown size
src/overlap.c(398) : error C2036: 'void *' : unknown size
src/overlap.c(399) : error C2036: 'void *' : unknown size
src/overlap.c(400) : error C2036: 'void *' : unknown size
src/overlap.c(401) : error C2036: 'void *' : unknown size
src/overlap.c(413) : warning C4028: formal parameter 7 different from declaration
src/overlap.c(502) : warning C4028: formal parameter 6 different from declaration
src/overlap.c(502) : warning C4028: formal parameter 7 different from declaration
```

Unfortunately it seems that void * is not allowed. Can you fix this issue? Or do I?

ExpHP:
I think I know what the problem is.  It looks like [pointer arithmetic on `void *` is actually a GCC extension](https://stackoverflow.com/questions/3523145/pointer-arithmetic-for-void-pointer-in-c).

The fix is to make sure all arithmetic is done on `char *` pointers.

I will submit a second PR to fix this.

PR #45, opened 2018-01-17
Fix void * arith on MSVC

ExpHP:
I compiled with `-Wpedantic` this time to make sure there were no GCC extensions accidentally being used.

cc @jochym, can you test this?  I don't have access to a computer with MSVC...


atztogo:
Thanks. I merged it to the master branch.

PR #47, opened 2018-02-22
cmake: use CMAKE_INSTALL_PREFIX

davydden:


atztogo:
I'm not good at cmake. Could you explain the difference for me?

davydden:
> Could you explain the difference for me?

well, you never install libraries into the source you are building from. You install them into `CMAKE_INSTALL_PREFIX`, see https://cmake.org/cmake/help/v3.0/variable/CMAKE_INSTALL_PREFIX.html

davydden:
also, `cmake` is out-of-source build, so I can build several versions from the same source and `cmake` script shall not pollute sources anyhow.

atztogo:
I see. Thanks.

PR #50, opened 2018-03-20
Fix debug build

serval2412:
/home/julien/projects/spglib/src/spglib.c: In function ‘set_dataset’:
/home/julien/projects/spglib/src/spglib.c:1188:19: error: ‘bravais’ undeclared (first use in this function)
   for (i = 0; i < bravais->size; i++) {
                   ^~~~~~~


atztogo:
Thanks!

PR #53, opened 2018-03-29
Fixing a typo in python/_spglib.c.

merkys:
As the typo is in the description string, I assume that this fix will not break anything.


atztogo:
Thanks!

PR #55, opened 2018-04-16
Update setup.py with long description and license

jochym:
The new pypi.io supports descriptions for the package.


atztogo:
Should it be one of them in the list: http://autopilot-docs.readthedocs.io/en/latest/license_list.html ?

merkys:
I find these license strings placed in ``classifiers``. However, in [scikit-bio](https://github.com/biocore/scikit-bio) I have found ``BSD-3-Clause`` indicated explicitly in their setup.py's ``license`` attribute.

PR #57, opened 2018-04-18
Adding missing metadata to use_setuptools==False branch of setup.py.

merkys:
There are two branches of the control flow and I assume that both of them should be as similar as possible.


atztogo:
Thanks.

PR #60, opened 2018-05-02
Small additional overlap.c optimization

colin-daniels:
Partially duplicate a version of `cel_is_overlap_with_same_type` into `overlap.c` to be used in `ovl_check_total_overlap`. Compiler macros are checked to determine if the actual inline keyword (or similar) can be reasonably assumed to work and are pretty conservative as far as I can tell. Otherwise, it falls back to no inlining and hoping that the compiler decides to do it on its own. 

Ideally, checks for support of C99 and other features would be done through something like autoconf, but the fact that there are essentially three separate build systems (autoconf, cmake, and python) makes it harder. That said, it is doable if needed.

On my machine I see a significant increase of performance for `spg_get_dataset` for large highly symmetric structures (80-100%), with a small gain running the unit tests (~5%).

### DIFF
--- a/src/overlap.c
+++ b/src/overlap.c
@@ -44,7 +44,21 @@
 /* 'a++' generalized to an arbitrary increment. */
 /* Performs 'a += b' and returns the old value of a. */
 #ifndef SPG_POST_INCREMENT
-#define SPG_POST_INCREMENT(a, b) (a += (b), a - (b))
+#define SPG_POST_INCREMENT(a, b) ((a) += (b), (a) - (b))
+#endif
+
+/* deal with inline */
+#if   defined(_MSC_VER)
+# define OVL_INLINE __forceinline
+#elif defined(__clang__)
+# define OVL_INLINE __inline__ __attribute__((__always_inline__))
+#elif defined(__GNUC__) && \
+    ( defined(__GNUC_STDC_INLINE__) || \
+     (defined(__STDC__) && (__STDC_VERSION__ >= 199901L)) || \
+     (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 2)) )
+# define OVL_INLINE __inline__ __attribute__((__always_inline__))
+#else
+# define OVL_INLINE /* empty */
 #endif
 
 /* Note: data_out and data_in MUST NOT ALIAS. */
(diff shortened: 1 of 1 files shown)


atztogo:
Thanks. I am not good at the tuning issue. So I can't evaluate it well but it seems the increase of difficulty of the maintenance is a little, i.e., if I am lost, returning to `cel_is_overlap_with_same_type` is reserved as a safe choice. Only my concern is `symprec * symprec` can underflow.  In the current algorithm, `symprec` is reduced when symmetry is not well found, which can hit easier to underflow if `symmprec * symprec` is used. Also, a user can choose 1e-9 or something like this, however I assume by my feeling users will not choose `symprec=1e-19`. So I expect `cartesian_norm_sq` to be not squared. Can this impact largely to the performance?

colin-daniels:
I tested with `sqrt` and it seems to take about ~14% more time with it for my test structure.

Otherwise, please correct me if I'm wrong but I don't think issues with machine epsilon should come into play here though since I checked and the only place the `cartesian_norm_sq` is used is in one comparison, not addition/subtraction. Much of the reason why I made my changes local to `overlap.c` and held off on doing any other supposedly "obvious" fixes was because I don't know where/how all the functions are used. 

That said, if you are worried about it being used elsewhere or something similar I can just put the code directly in the static `has_overlap` function in `overlap.c` so it's not used elsewhere, or just put the `sqrt` in since the performance hit is only about ~14% on my admittedly unusual structure (space group 230 with 960 atoms). Let me know what you prefer and if there are any other concerns.

atztogo:
Maybe this way,
```
if (cartesian_norm_sq(lattice, v_diff) < (symprec * symprec)) {
  return 0;
} else {
  return 1;
}
```

colin-daniels:
Ah, I understand your concern now, thanks for explaining. Doing some longer benchmarking I have the following for my test structure and calling `spg_get_dataset`:

```
code     speedup
-------- -------
original  1.00x
square    1.86x
sqrt      1.63x
sq + e    1.85x
```
Honestly, since it's working as it is now with square root, it may not be worth it to do "perfectly" since I think it depends on the user's input of lattice vectors for how bad the error can be.

Actually looking at the error for "identical" points, its fairly convoluted where if you allow a maximum difference of `ϵ` in lattice coordinates due to periodicity issues (e.g. wrapping from `1 - ϵ` -> `0`) then it ends up being something like
```
error(delta) = (±ϵ, ±ϵ, ±ϵ)
error(cartesian_norm_sq) = ϵ^2[(ax + bx + cx)^2 + (ay + by + cy)^2 + (az + bz + cz)^2]
```
where `ax, bx, ...` are the lattice vector components. So in the worst case scenario of an atom on the periodic boundary and something like lattice vectors of length `1000` or something, it works out to around `4e-24`, which is still less than epsilon but yeah it may not be worth it to try to "perfectly" figure it out.

That said, if you think it's worth it something like `cartesian_norm_sq(lattice, v_diff) < symprec^2 + DBL_EPSILON` should be conservative enough, it just effectively limits `symprec` to around `1e-8` for users. As for implementation, `DBL_EPSILON` from `<float.h>` is accessible on all POSIX or C89 and later systems, so that should work without any messing around with macros or compatibility.

atztogo:
I wanted to mean
```
cartesian_norm_sq(lattice, v_diff) < symprec^2 + DBL_EPSILON
```
is similar to
```
!(cartesian_norm_sq(lattice, v_diff) > symprec^2)
```

In spglib, `symprec` is the distance tolerance (measure) in Euclidean coordinates not in lattice coordinates.

colin-daniels:
I've just gone ahead and switched back to the `sqrt` method since it has been working, let me know if there are any different issues.

atztogo:
Thanks. I merged your commits to the develop branch. This will be merged to the master branch together with another pull request that will be ready to be merged to the master branch.

PR #70, opened 2019-01-29
Include <stddef.h> for size_t

heirecka:
Otherwise building the tests fails for me with: "In file included
from test.c:1: spglib.h:382:3: error: unknown type name 'size_t'".


atztogo:
Yes, you are correct! Merged. Thanks!

PR #76, opened 2019-03-21
Access some functions by International Table spacegroup name

g5t:
In situations where the unit cell is not fully populated but where the correct spacegroup is still known -- e.g., users of [SpinW](https://github.com/spinw/spinw) often omit non-magnetic atoms from their lattice definitions -- it could be useful to have access to spglib's database of symmetry operations or to let it create an irreducible reciprocal grid.
To that end, I've added C and Python functions to access symmetry information by matching a supplied string against one of the three International Table spacegroup strings stored in the `SpacegroupType` database, `international`, `international_long`, or `international_short` to find a (potentially not-unique) Hall number.

I've also added functions to access a spacegroup's pointgroup rotation matrices to enable identification of _which_ rotation (and of which order) is necessary to map points on the reciprocal grid to their equivalent irreducible grid point. With this information one can, e.g., calculate a vector or tensor quantity for each irreducible grid point and transform it to any non-irreducible grid point.
If there's any interest in it, I could add functions to perform the identification inside of spglib, potentially during a call to `*get_ir_reciprocal_mesh` if breaking its API is OK in a future release.

Finally, I have made a small change to `CMakeLists.txt` which should allow for the C library to be built successfully on Windows systems using native Visual Studio compilers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,11 @@ set(SOURCES ${PROJECT_SOURCE_DIR}/src/arithmetic.c
 add_library(symspg SHARED ${SOURCES})
 set_property(TARGET symspg PROPERTY VERSION ${serial})
 set_property(TARGET symspg PROPERTY SOVERSION ${soserial})
-install(TARGETS symspg LIBRARY DESTINATION ${CMAKE_INSTALL_PREFIX}/lib)
+if (WIN32)
+	install(TARGETS symspg RUNTIME DESTINATION ${CMAKE_INSTALL_PREFIX}/lib)
+else()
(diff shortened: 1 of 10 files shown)


atztogo:
It's a big pull request.

Support of software is killing me by reducing my time to do scientific research. I am struggling to reduce time to support software, for which the over all design of code, especially the design of API is very important. I don't want to suffer from backward compatibility etc. So those features that can be separated from the core of spglib are expected to be maintained out of spglib. For this purpose,

1) I want to minimize the number of public functions.
2) I plan to separate kpoints handling out of spglib.

Do you think your added features can be put out of spglib. What is minimum API required to realize what you want to do?

atztogo:
No reply for many months. So I will close this PR. But please re-open whenever you like.

PR #77, opened 2019-05-29
Changed strcpy() calls to strncpy() to avoid buffer overflows

krefson:
Dear Atsushi,

This patch fixes a bug which caused a crash within the Fortran interface of CASTEP, and which causes the spglib test "make check" to fail.  The "strcpy" calls were in some instances overrunning the output buffer and causing mamory corruption.  The fix is to change to "strncpy".

Keith Refson

### DIFF
--- a/src/arithmetic.c
+++ b/src/arithmetic.c
@@ -150,7 +150,7 @@ int arth_get_symbol(char symbol[7], const int spgroup_number)
   }
 
   arth_number = arithmetic_crystal_classes[spgroup_number];
-  strcpy(symbol, arithmetic_crystal_class_symbols[arth_number]);
+  strncpy(symbol, arithmetic_crystal_class_symbols[arth_number], sizeof symbol - 1);
   for (i = 0; i < 6; i++) {
     if (symbol[i] == ' ') {symbol[i] = '\0';}
   }
--- a/src/pointgroup.c
+++ b/src/pointgroup.c
@@ -460,8 +460,8 @@ Pointgroup ptg_get_pointgroup(const int pointgroup_number)
 
   pointgroup.number = pointgroup_number;
   pointgroup_type = pointgroup_data[pointgroup_number];
-  strcpy(pointgroup.symbol, pointgroup_type.symbol);
-  strcpy(pointgroup.schoenflies, pointgroup_type.schoenflies);
+  strncpy(pointgroup.symbol, pointgroup_type.symbol, sizeof pointgroup.symbol - 1);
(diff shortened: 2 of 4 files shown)


atztogo:
Thanks. Merged to the develop branch. Do you need to be released? I replaced `sizeof(symbol) - 1` by the explicit numbers to avoid gcc warnings for type. I think for C, `-1` is not to be there, so effectively I wrote the numbers of `sizeof(symbol)`. Can you check if this works for your Fortran interface?

krefson:
Dear Atsushi,

Thank you.  I believe the "-1" *is* necessary, to accommodate "n" character plus the trailing NULL character.  So if the largest allowable string is 8 characters it should be declared as length [9] to leave space for the terminating NULL. In that case "strncpy(b,a,8)" will not write where it should not, but it could leave a string with no terminating NULL, which could cause trouble when any further use is made of it (such as printing).

As I wrote in the comment, the bug shows up in the "test_spglib" tests, particularly if run under "valgrind".  But I will also test it from the Fortran - what happened there was that the "get_dataset" interface was attempting to copy "std_types" from the C struct to the Fortran type, but as "n_std_atoms" had the value in the region of 9e8 this caused a segfault.  Clearly this is a case of a buffer overrun corrupting later data. It took me a while to identify the root cause!

What I did not write in the pull request is that the behaviour depends on the particular compiler version.  Gfortran 7.4.0 showed symptoms but 8.2.0 does not!  Presumably the memory layout is slightly different and the consequences.

In any case, thank you for accepting my bug report.

Keith Refson

On 29/05/2019 15:12, Atsushi Togo wrote:

Thanks. Merged to the develop branch. Do you need to be released? I replaced sizeof(symbol) - 1 by the explicit numbers to avoid gcc warnings for type. I think for C, -1 is not to be there, so effectively I wrote the numbers of sizeof(symbol). Can you check if this works for your Fortran interface?

atztogo:
I think we agree that `char foo[8]` can contain 7 characters + `\0`. Now I want to copy char[] variables including `\0`. As far as I read the manual of `strncpy`, this copies also `\0`. So I think `-1` is unnecessary. What I want to mean is shown in the following C code. But I may misunderstand what you mean...

```c
#include <stdio.h>
#include <string.h>
int main() {
  char src[7] = "foobar";
  char dst[7];
  printf("(src) %s\n", src);
  printf("(src) is null? %d\n", src[6]);

  strncpy(dst, src, 7);

  printf("(dst) %s\n", dst);
  printf("(dst) is null? %d\n", dst[6]);
}
```

The output is as below,
```
% clang test.c
% ./a.out
(src) foobar
(src) is null? 0
(dst) foobar
(dst) is null? 0
```

Issue #84, opened 2020-02-13
symspg is missing libm in target_link_libraries


StefanBruens:
When compiling with the --no-undefined linker flag, linking fails:

```
/usr/bin/cc -fPIC -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -g -DNDEBUG -O2 -g -DNDEBUG -flto=auto -Wl,--as-needed -Wl,--no-undefined -Wl,-z,now -shared -Wl,-soname,libsymspg.so.1 -o libsymspg.so.1.14.1 CMakeFiles/symspg.dir/src/arithmetic.o CMakeFiles/symspg.dir/src/cell.o CMakeFiles/symspg.dir/src/debug.o CMakeFiles/symspg.dir/src/delaunay.o CMakeFiles/symspg.dir/src/determination.o CMakeFiles/symspg.dir/src/hall_symbol.o CMakeFiles/symspg.dir/src/kgrid.o CMakeFiles/symspg.dir/src/kpoint.o CMakeFiles/symspg.dir/src/mathfunc.o CMakeFiles/symspg.dir/src/niggli.o CMakeFiles/symspg.dir/src/overlap.o CMakeFiles/symspg.dir/src/pointgroup.o CMakeFiles/symspg.dir/src/primitive.o CMakeFiles/symspg.dir/src/refinement.o CMakeFiles/symspg.dir/src/site_symmetry.o CMakeFiles/symspg.dir/src/sitesym_database.o CMakeFiles/symspg.dir/src/spacegroup.o CMakeFiles/symspg.dir/src/spg_database.o CMakeFiles/symspg.dir/src/spglib.o CMakeFiles/symspg.dir/src/spin.o CMakeFiles/symspg.dir/src/symmetry.o 
[   17s] /usr/lib64/gcc/x86_64-suse-linux/9/../../../../x86_64-suse-linux/bin/ld: /tmp/libsymspg.so.1.14.1.Ny5QwT.ltrans0.ltrans.o: in function `get_angle':
[   17s] /home/abuild/rpmbuild/BUILD/spglib-1.14.1/src/symmetry.c:858: undefined reference to `acos'
[...]

PR #86, opened 2020-02-13
Fix missing libm in target_link_libraries

StefanBruens:
Linking with --no-undefined linker flag fails otherwise, as e.g.
`sqrt` is not defined.

Fixes issue #84.


atztogo:
Thanks. Merged to develop branch.

atztogo:
This is closed because fixed.

PR #89, opened 2020-03-15
Rust bindings

seatonullberg:
This is the initial release of my [Rust bindings to spglib](https://crates.io/crates/spglib). The additions are all in the `rust` directory which contains auto generated raw bindings as well as my more idiomatic interface.

Included Features:
* `Cell` object which contains the lattice, positions, and types of a crystalline cell
  * Wraps `spg_standardize_cell`, `spg_delaunay_reduce`, and `spg_niggli_reduce`
* `Dataset` object which closely wraps `SpglibDataset`
  * Can be initialized as `new` or `with_hall_number` by wrapping `spg_get_dataset` and `spg_get_dataset_with_hall_number`
* `Spacegroup` object which closely wraps `SpglibSpacegroupType`
* `hall_number_from_symmetry` function  which closely wraps `spg_get_hall_number_from_symmetry`
* Unit tests and documentation of completed features


rust/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "spglib"
+version = "0.1.2"
+authors = ["Seaton Ullberg <seatonullberg@gmail.com>"]
+edition = "2018"
+license = "BSD-3-Clause"
+description = "Rust-y bindings to spglib"
+repository = "https://github.com/seatonullberg/spglib"
```
merkys: Shouldn't this URL be changed to `https://github.com/atztogo/spglib` prior to merge?

rust/spglib-sys/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "spglib-sys"
+version = "0.1.0"
+authors = ["Seaton Ullberg <seatonullberg@gmail.com>"]
+edition = "2018"
+license = "BSD-3-Clause"
+description = "Raw bindings to spglib"
+repository = "https://github.com/seatonullberg/spglib"
```
merkys: This URL here too should point to this repository.

rust/Cargo.toml
seatonullberg: Good catch, fixed in https://github.com/atztogo/spglib/pull/89/commits/ec957715a6d3800a2a0a4a6bca6a489474b982c1

rust/spglib-sys/Cargo.toml
seatonullberg: Also fixed in https://github.com/atztogo/spglib/pull/89/commits/ec957715a6d3800a2a0a4a6bca6a489474b982c1

seatonullberg:
@merkys To build from source you will need Rust's package management tool [Cargo](https://doc.rust-lang.org/cargo/). 
```bash
cd spglib/rust/  # enter the project directory
cargo test  # confirm that all tests pass
cargo build --release  # compile the project with optimizations
cargo doc --open  # optionally, generate and view the documentation in a local browser
```
However, I have already published the package as an alpha release (using my branch as source) on the official Rust package index [crates.io](https://crates.io/crates/spglib) so users can simply specify `spglib = "0.1.2"` as a dependency in their `Cargo.toml` files.

seatonullberg:
@atztogo Thank you for that catch. I have modified the behavior in https://github.com/atztogo/spglib/pull/89/commits/458aa755cc43abe873e76dfe92280603d924db2f to link directly to the user's installed version of spglib.h.

atztogo:
Thanks @seatonullberg , @merkys. Can I merge?

merkys:
Just a question: wouldn't it make more sense to have the same version number through all the components of spglib (I.e., taken from `src/version.h`?)? Otherwise it may be difficult to keep track of all of them.

atztogo:
Supporting information of @merkys 's question , how it is done in python wrapper.

For `spglib.__version__`,
https://github.com/atztogo/spglib/blob/master/python/spglib/__init__.py

For the package
https://github.com/atztogo/spglib/blob/f8ddf5b2b242c6f1b8cf090545e236a48dffe06b/python/setup.py#L108

seatonullberg:
I am not sure this is possible due to Rust's [Cargo.toml manifest format](https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field). Since it is not code, (unlike Python's setup.py) it cannot dynamically import data. However, I do expose the `spg_get_major_version`, `spg_get_minor_version`, and `spg_get_micro_version` functions in my interface so end users will always be able to determine which underlying version of spglib is being used.

merkys:
@seatonullberg could we generate `Cargo.toml` from `Cargo.toml.in` (for example) using Makefile or some other tool? I would really prefer metainfo of Rust bindings have the same version as spglib itself, especially if it stays in the same repository (just as Python bindings do).

seatonullberg:
Yes, I think that would be a viable solution. To clarify, this would mean updates to the bindings will be released on the same schedule as the main library? I initially thought the bindings would be free to change independent of the source, but I can see how this could be more organized.

merkys:
@seatonullberg:
> To clarify, this would mean updates to the bindings will be released on the same schedule as the main library?

I would prefer so. Changes in the main library should affect the bindings too.

seatonullberg:
@merkys @atztogo Thank you both for the advice. That fix will take some more time so I will close for now and resubmit once I've implemented the changes.

merkys:
@seatonullberg Thanks a lot for the interesting PR. Will gladly review the updated version once it's done!

atztogo:
Thanks @seatonullberg and @merkys .
 
Great. Merged.

@merkys 's advise will be useful in the future once many users will start to use @seatonullberg's rust binding. We are simply happy if @seatonullberg considers about versioning. I think the first priority is to fulfill @seatonullberg 's demand. I may request to write some documentation for other users, but it is just for your voluntee.

seatonullberg:
@atztogo Thanks for the merge! I have already written some documentation which is hosted [here](https://docs.rs/spglib/0.1.2/spglib/index.html). It does not match the style of the existing docs but maybe you could link to it in the interim.

atztogo:
@seatonullberg , I will make a link from spglib web site. But the URL you gave is specific for a version. Is there any better URL so that I will not need to update?

seatonullberg:
@atztogo This link should point to the most recent version. 

https://docs.rs/crate/spglib/

Since removing the the spglib.h file, the documentation builder has failed to build the most recent version, but the prior ones are still available. This is due to the fact that spglib is not installed in the container environment which [docs.rs](https://docs.rs/) uses to compile the package and build the documentation. I have [filed an issue](https://github.com/rust-lang/docs.rs/issues/654) and am waiting for a response.

atztogo:
I put a link in the sphinx source of the documentation at the commit 18b4db4.

Issue #90, opened 2020-04-02
Problem with equivalent atoms


lauri-codes:
First, thank you for creating this superb library! I have used it quite a lot and it has worked wonders.

Just recently I noticed that there was an inconsistency with `equivalent_atoms` that I could not figure out. It may be entirely due to me not understanding what is going on but I hope you could shed a light on what is happening.

When running the following code

```python
import spglib

positions = [
    [0.00000000e+00, 0.00000000e+00, 0.00000000e+00],                                                                                    
    [5.00000783e-01, 5.00000105e-01, 5.00000504e-01],
    [2.57675031e-01, 2.57675031e-01, 0.00000000e+00],
    [7.42324606e-01, 7.42324606e-01, 0.00000000e+00],
    [7.42441461e-01, 2.57559427e-01, 4.18791129e-02],
    [4.57816602e-01, 5.42184286e-01, 2.42498450e-01],
    [4.13846169e-02, 9.58616271e-01, 2.57474206e-01],
    [7.57373910e-01, 2.42626978e-01, 4.58800327e-01],
    [2.42314486e-01, 2.42313809e-01, 5.00000504e-01],
    [7.57685827e-01, 7.57685150e-01, 5.00000504e-01],
    [2.42627655e-01, 7.57373233e-01, 5.41200681e-01],
    [9.58616953e-01, 4.13839350e-02, 7.42525529e-01],
    [5.42183696e-01, 4.57817192e-01, 7.57501282e-01],
    [2.57560109e-01, 7.42440779e-01, 9.58120623e-01],
    [9.99745256e-01, 5.00077432e-01, 2.50229920e-01],
    [4.99921738e-01, 2.53959000e-04, 2.50230128e-01],
    [5.00078114e-01, 9.99744574e-01, 7.49769816e-01],
    [2.54636000e-04, 4.99921061e-01, 7.49770880e-01],
    [5.00000444e-01, 5.00000444e-01, 0.00000000e+00],
    [3.38500000e-07, 9.99999661e-01, 5.00000504e-01],
]
numbers = [28, 28, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 38, 38, 38, 38, 52, 52]
lattice = [
    [-3.93057, -3.994236, -0.010812],
    [3.93057, -3.994236, 0.010812],
    [-0.030735, 0.0, 7.861732]
]
cell = (lattice, positions, numbers)

# Get symmetry dataset
[...]

PR #91, opened 2020-04-04
Added crystallographic orbits

lauri-codes:
Added the crystallographic orbits to the dataset as discussed in issue #90. Should be a relative straightforward as only new functionality is added and the old is unaffected. Feel free to modify as you see fit.

### DIFF
--- a/python/_spglib.c
+++ b/python/_spglib.c
@@ -217,7 +217,7 @@ static PyObject * py_get_dataset(PyObject *self, PyObject *args)
 
   PyObject *array, *vec, *mat, *rot, *trans, *wyckoffs, *equiv_atoms;
   PyObject *site_symmetry_symbols, *mapping_to_primitive;
-  PyObject *std_lattice, *std_types, *std_positions, *std_mapping_to_primitive, *std_equiv_atoms;
+  PyObject *std_lattice, *std_types, *std_positions, *std_mapping_to_primitive, *crystallographic_orbits;
   PyObject *std_rotation;
 
   int i, j, k, n;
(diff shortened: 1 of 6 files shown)


atztogo:
I thought what you want is the symmetrically equivalent atoms in the input crystal without considering symmetry breaking by the input lattice. But as far as I read your PR, what you want is the symmetrically equivalent atoms in standardized crystal structure. Am I correct? Then now I see your naming `std_equivalent_atoms` is probably better than `crystallographic_orbits`.

atztogo:
`wyckoffs` are given for the atoms of the input crystal structure but not the standardized crystal structure. So I thought as I wrote above.

lauri-codes:
Yes, I want the symmetrically equivalent atoms in the standardized cell, sorry if this was unclear. Should I change the names to std_equivalent_atoms?

atztogo:
I'm glad that we can communicate using code.

I see. Then we should have `std_wyckoffs`, too. We can add three new members, `std_wyckoffs`, `std_crystallographic_orbits`, and `crystallographic_orbits` during v1.x. I will merge and modify your code to the develop branch.

lauri-codes:
Ok, perfect!

atztogo:
I spent some time to think about how it should be done. I merged your PR and now `crystallographic_orbits` is accessible, but not by your definition. I implemented as I misunderstood about it. The length of `crystallographic_orbits` is the number of atoms in the input crystal structure, but not the standardized unit cell. In your example, these are the same number though. I chose this definition because I don't want to increase the specification if possible to avoid increasing maintenance cost. 

`crystallographic_orbits` and `equivalent_atoms` in standardized unit cell must be the same, and they can be obtained via `mapping_to_primitive`, `std_mapping_to_primitive`, and `crystallographic_orbits` in principle. So with a little more code, you can get those information. But it requires like 10 lines in python code. So I am wondering how I can provide these information currently.

atztogo:
The following is the python code to obtain `std_crystallographic_orbits`. This code can be much simple if I am clever.

```python
    equiv_atoms_to_primitive = {i: j for i, j in zip(
        symmetry_dataset['mapping_to_primitive'][::-1],
        symmetry_dataset['crystallographic_orbits'][::-1])}
    equiv_atoms_prim = [equiv_atoms_to_primitive[i]
                        for i in range(len(equiv_atoms_to_primitive))]
    equiv_atoms_std = [equiv_atoms_prim[i]
                       for i in symmetry_dataset['std_mapping_to_primitive']]
    std_crystallographic_orbits = {}  # atom indices are those in input cell
    for i, j in enumerate(equiv_atoms_std):
        if j not in std_crystallographic_orbits:
            std_crystallographic_orbits[j] = i
    # atom indices are mapped to those in the standardized cell.
    std_crystallographic_orbits = [std_crystallographic_orbits[i]
                                   for i in equiv_atoms_std]
```

atztogo:
@lauri-codes , is this OK for you?

lauri-codes:
Hi @atztogo,
This is perfectly fine. Thanks a lot!

atztogo:
OK. I will release soon a new version including this PR. Thanks for your PR. I will close this PR.

PR #95, opened 2020-04-28
do not link to libm for msvc build

bodomartin:
now compiles with msvc 2019 16.5

(I have noticed that the tests (make check, which runs spglibtest) fail, both with gcc and msvc)


atztogo:
If I merge to master, it is automatically released. So if you don't mind, could you send PR to develop? Sorry for this inconvenience.

atztogo:
This was merged in PR #96. So I close this PR. Thanks.

PR #110, opened 2020-09-09
Migrate rust interface to spglib-rs repository

seatonullberg:
All of the rust interface functionality has been migrated to [its own repository](https://github.com/spglib/spglib-rs) within the spglib organization. Versioning is now consistent with the C library and documentation is hosted [here](https://docs.rs/spglib).


atztogo:
Thanks! Merged.

PR #117, opened 2021-02-06
Added standardize_cell to f08 interface

Jonas-Finkler:
Hi, 
I was trying to use the standardize_cell function from a fortran code and realized the interface was not properly implemented. 
I added spg_standardize_cell and spgat_standardize_cell to the public interface and fixed the return type in the c bindings.


atztogo:
Thanks. It is indeed a bug. Merged to develop.

PR #119, opened 2021-02-26
install fwrapper with cmake

dmt4:
Having the fortran wrapper installed would be a convenience. This is the main reason for the pr. I also fixed some warnings about about the cmake build not using omp flags and initialisation of some var.

It does not look like the f wrapper is tested in any way, maybe tests could be ported from c or python?


atztogo:
Thanks.
You are correct. No test is made for fortran wrapper itself, though it's done via python wrapper. So I believe for C and python, they are well tested. Writing equivalent test for fortran to that for python is not a small job, but it would be good to be done for at least some.

PR #120, opened 2021-03-30
fix(delaunay): fix error in calculating Delaunay change-of-basis matrix

site-g:
[https://github.com/spglib/spglib/blob/develop/src/primitive.c#L418](url) the argument `min_lattice` and `lattice` point to the same memory-address. This makes [https://github.com/spglib/spglib/blob/develop/src/delaunay.c#L135](url) always equal to identity matrix rather than change-of-basis matrix.

remove redundent assignment in cell.c.

### DIFF
--- a/src/cell.c
+++ b/src/cell.c
@@ -72,8 +72,6 @@ Cell * cel_alloc_cell(const int size)
     return NULL;
   }
 
-  cell = NULL;
-
   if ((cell = (Cell*) malloc(sizeof(Cell))) == NULL) {
     warning_print("spglib: Memory could not be allocated.");
     return NULL;
--- a/src/delaunay.c
+++ b/src/delaunay.c
@@ -92,7 +92,9 @@ static int delaunay_reduce(double red_lattice[3][3],
   int i, j, attempt, succeeded;
   int tmp_mat_int[3][3];
   double volume;
-  double tmp_mat[3][3], basis[4][3];
+  double orig_lattice[3][3], tmp_mat[3][3], basis[4][3];
+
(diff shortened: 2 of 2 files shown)


atztogo:
Could you explain why `mat_copy_matrix_d3(orig_lattice, lattice);` is needed?

site-g:
- `red_lattice` and `lattice` both point to `prim_lattice`
https://github.com/spglib/spglib/blob/8e8ecd21e8239501177d59e0a37feae417be44e1/src/primitive.c#L418
https://github.com/spglib/spglib/blob/8e8ecd21e8239501177d59e0a37feae417be44e1/src/delaunay.c#L88-L90

 - `red_lattice` changes to delaunay basis, so do `lattice`
 https://github.com/spglib/spglib/blob/8e8ecd21e8239501177d59e0a37feae417be44e1/src/delaunay.c#L112-L116

 - It should calculate `red_lattice`<sup>-1</sup>·`lattice_before_change` == _**Q**<sub>pi</sub>_, but in fact it is `red_lattice`<sup>-1</sup>·`lattice_after_change` == `red_lattice`<sup>-1</sup>·`red_lattice` == _**I**_
 https://github.com/spglib/spglib/blob/8e8ecd21e8239501177d59e0a37feae417be44e1/src/delaunay.c#L134-L135

 - So we need to copy the origin cell

atztogo:
Yes, I see. You are correct! Thanks.

This check is useless.
https://github.com/spglib/spglib/blob/8e8ecd21e8239501177d59e0a37feae417be44e1/src/delaunay.c#L134-L141

atztogo:
Merged. Thanks!

PR #124, opened 2021-05-17
fix(Wyckoff): fix error in wyckoff database

site-g:
The wyckoff position of `P 1 1 b` is inconsitent with International Tables for Crystallography (2016). Vol. A, Space group 7, pp. 215. The one in ITA seems more credible.


atztogo:
@site-g, thanks indeed!

PR #133, opened 2021-08-18
openmp on windows/msvc needs signed type for loop var i

bodomartin:
Hello,

the current master fails to build on windows (msvc/clang). because for the openmp pragma the loop
var 'i' in src/kpoint.c must be signed, i.e. ptrdiff_t instead of size_t.

These are the only changes in the pull request


atztogo:
Thanks.
Diff (https://github.com/spglib/spglib/pull/133/files) is large, so I may modify the code only the lines of the issues.
I am considering replacing size_t to long. So can long type int work for your msvc/clang?

bodomartin:
yes sorry about the size of the diff, thinking about it it is probably line endings (I copied the file from wondows after modification)

I think the most compatible way would be ptrdiff_t for signed and size_t for unsigned - but long would do as well. I can also try to prepare a minimal diff

bodomartin:
I have tried to make the patch 'minimal'  which it is, i.e. when excluding the "space changes" in gitk it is only ~ 6 lines - the commit does remove extra spaces at the end of the lines.

atztogo:
Thanks. I am a little bit afraid of using `ptrdiff_t` because it might break windows binary build in conda-forge. So if your build system allowed, can it be changed to `long`? If you are fine, I will do it in merging your PR.

bodomartin:
ok done changed the signed type to long

atztogo:
Thanks again. Pushing to master means the official release and triggers automatic deploy system. So this PR was merged to the develop branch, https://github.com/spglib/spglib/commit/c939a36a6f0e44b2e0a53e55c7c72162dcf502bd.

atztogo:
I will close this PR because already merged.

PR #137, opened 2021-10-08
Clarify primitive setting choice

thchr:
The primitive setting choice described in the documentation is consistent with the CDML primitive setting choice, as can be confirmed e.g. by comparison with [Table 2 of this paper](https://doi.org/10.1107/S205327331303091X) [which give (P⁻¹)ᵀ] or by comparison with Tables 1.5.4.1 and 1.5.4.2 of the International Table of Crystallography, Vol. B, 2nd edition.

I figured it might be helpful to explicate this (e.g., so users can see if the setting is compatible with settings used by other libraries or reference works).

### DIFF
--- a/doc/definition.rst
+++ b/doc/definition.rst
@@ -348,7 +348,7 @@ as follows:
 
 The choice of transformation matrix is a matter of convention. The transformation
 matrices above transform from a conventional ITA basis to the primitive
-setting adopted in the reference work of Cracknell, Davies, Love, and Miller
+setting adopted in the reference work of Cracknell, Davies, Miller, and Love
 (CDML). Other primitive setting choices exist: of note, the CDML choice is not
 the primitive setting choice made in the International Tables of Crystallography.
 


thchr:
Also not a crystallography specialist here, unfortunately, but had a need to find out which convention was being used.

Regarding "tI" and ITB: Tables 1.5.4.1 and 1.5.4.2 actually imply the same matrix for "tI" as it does for "oI" and "cI". The slightly tricky thing about the tables is that it is not each table individually that gives a matrix, but their combination (since the relevant transformation is from **k** to **k**<sub>p</sub>).
Summarizing, Table 1.5.4.1 gives a mapping **k** = **P**<sub>1</sub><sup>T</sup>**k**<sub>a</sub> and Table 1.5.4.2 gives a mapping **k**<sub>p</sub> = **P**<sub>2</sub><sup>T</sup>**k**<sub>a</sub>. The mapping we want is **k**<sub>p</sub> = **P**<sup>T</sup>**k** so we have to reconstruct **P** via **P**<sup>T</sup> = **P**<sub>2</sub><sup>T</sup>(**P**<sub>1</sub><sup>T</sup>)<sup>-1</sup>, i.e. **P** = **P**<sub>1</sub><sup>-1</sup>**P**<sub>2</sub>. If you use these rules, you can verify that "oI" and "cI" has the same **P** matrix as "tI" (I just double checked myself). So, although more laborious, it is consistent with Aroyos paper.

Unfortunately I don't have CDML handy right now (I had it in my office for the last year and a half, but finally gave it back to the library last week... unlucky). I think Harold Stokes' book on isotropy groups might have a table regarding setting choices as well that would likely also include the CDML convention - but I think ITB and Aroyos paper should probably be enough confirmation already.

atztogo:
I assume main users of spglib are not specialists of crystallography. The clarifying the primitive setting choices may be useful for some users, but may confuse more users, for example, the correspondence between matrices in direct and reciprocal spaces is not intuitive for many users. So I think, if we write it, we should provide more detailed information like what we discuss here.
On CDML, if we don't confirm it, we can't write it on the documentation. CDML is rarely found in Japan. I found it exists in the university near my place, but currently it is not open for non-university members because of COVID.

atztogo:
Thanks again for your valuable comments. A university in Japan could ship CDML to me, and now I got it! It is our job to go to the original source if possible.

Those conventions are not easy. I chose the transformation matrices from conventional unit cells to the primitive cells by the simplicity of the shapes of the matrices. I don't remember correctly, but I think I am most influenced by Grosse-Kunstleve's papers (references found at https://arxiv.org/abs/1808.01590) and probably some other text books. I tried to understand the Aroyo's papers, but as you could see, I did not understand those very well.

I think your information is valuable to dedicate for one page of spglib web site. I would appreciate it if you could spare time for writing anything you consider useful for practical users, but not necessary to be too rigorous. If crystallographers will be interested in it, they would improve it in their sense. I know spglib is not rigorous crystallographically. But I need to make it because I could not find a convenient library with an acceptable open source license for me, and it is simply my surprise that many people now use spglib. If someone else needs to make it more rigorous, I hope they will do it.

atztogo:
@thchr, I'm sorry being late to reply. @lan496 and I checked CDML, and we found they are not equivalent. The transformation matrices for rhombohedral are different. So unfortunately you commit is not very correct.

Table. 1.5.4.2 in ITB and Table 2 in Acta Cryst. (2014). A70, 126-137 by Aroyo et al. are similar but may be different for tI.

Tables 1.5.4.1 in ITB or Table 3 in the above paper may not be consistent because the choice of the conventional basis vectors in reciprocal space may not be unique.

We consider Table 2 in Aroyo's paper is easiest to compare with the spglib documentation and they are equivalent at the transformation matrix level.

thchr:
@atztogo I'm doubly sorry for replying so slowly, no less than half a year later.

I got CDML sent to my library again, and was checking it against the matrices listed in spglib just now. Indeed, you are correct that the 'trigonal, R' transformation matrix (of table 3.3 in CDML), i.e. for trigonal rhombohedral lattices, apparently does not match up with what is in spglib's documentation - or for that matter, the "standard" primitive transformation matrices of ITA/ITB.

For posterity, the differences are:
image

The choice made by CDML is what ITA calls the 'triple hexagonal cell R<sub>3</sub>, reverse setting'; whereas the "standard" choice from ITA - and the choice made by spglib also - is the 'triple hexagonal cell R<sub>1</sub>, obverse setting'.

Aside from this difference in the rhombohedral case, the rest of the matrices are consistent with CDML's.

Still, in light of this discrepancy, I agree it would be misleading to call this fully consistent with the CDML choices: I could change the PR to say that this is consistent with the standard primitive basis described in ITA (e.g., with a reference to associated tables) and that it is additionally consistent with CDML, apart from the rhombohedral setting?

thchr:
Also for posterity, here's the relevant page from CDML:

atztogo:
I agree with your direction. CDML is not easily accessible. If we write something wrong, we will confuse people. But if the information to be documented is not wrong and not misleading, more information must be useful to be documented.

PR #153, opened 2022-04-11
CMake improvements

e-kwsm:
* Remove `-Wall` from `CMAKE_C_FLAGS_RELEASE` since it is not portable
* Use `add_definitions` for debugging macros
* Make OpenMP mandatory if `USE_OMP`
* ~~Use `cmakedefine` for `SPGLIB_(MAJOR|MINOR|MICRO)_VERSION`~~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,11 +6,12 @@ project(spglib C)
 option(WITH_Fortran "enable f08 interface" OFF)
 
 set(CMAKE_MACOSX_RPATH 1)
-set(CMAKE_C_FLAGS_RELEASE "-Wall -O2")
-set(CMAKE_C_FLAGS_DEBUG "-g -DSPGDEBUG -DSPGWARNING")
 if(NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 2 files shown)


e-kwsm:
https://github.com/spglib/spglib/blob/911fe7129915d760a1efb5099219ef4383c558db/CMakeLists.txt#L2

CMake 2 is relatively old; how about bumping to 3?

atztogo:
> https://github.com/spglib/spglib/blob/911fe7129915d760a1efb5099219ef4383c558db/CMakeLists.txt#L2
> 
> CMake 2 is relatively old; how about bumping to 3?

I agree.

atztogo:
Honestly I am not good at cmake.

`version.h` is read from `python/setup.py`. The aim is to write version number only at one place. In principle, the location of the version number written can be anywhere. But in `src` may be reasonable. How do you think @e-kwsm?

e-kwsm:
Thank you for your quick response and review. I understand your point and have reverted the changes related to `version.h`.

atztogo:
Merged. Thank you very much @e-kwsm!

PR #154, opened 2022-04-22
Fix dead links

lan496:
Dr. Seto's page for Hall symbols seems to be moved.


atztogo:
Merged. Thanks @lan496!

atztogo:
Updated http://spglib.github.io/spglib/.

Issue #150, opened 2022-03-26
Return spin flip information for magnetic symmetry in python interface


jaemolihm:
Hi, thanks a lot for developing and maintaining this very nice package!

From the comments in the python code, it seems that magnetic symmetry with non-collinear magnetism is supported. But, `spglib.get_symmetry` does not give an information about whether the symmetry involves time reversal or not. For example, in the following code, first four symmetries do not have time reversal, and the others do.

1. Is this information available in other python interface function?
2. Even if this information is available in other interface function, it would be useful to have this information in the output of `spglib.get_symmetry`.

```python3
import spglib

lat = [[1, 0, 0], [0, 2, 0], [0, 0, 3]]
pos = [[0., 0., 0.], [0., 0., 0.5]]
num = [0, 0]
magmom = [[0., 0., 1.], [0., 0., -1.]]
cell = (lat, pos, num, magmom)

print(spglib.get_symmetry(cell).keys())
print(spglib.get_symmetry(cell)["translations"])
```
Output:
```
dict_keys(['rotations', 'translations', 'equivalent_atoms', 'primitive_lattice'])
[[-0.  -0.  -0. ]
 [-0.  -0.  -0. ]
 [-0.  -0.  -0. ]
 [-0.  -0.  -0. ]
 [-0.  -0.   0.5]
 [-0.  -0.   0.5]
 [-0.  -0.   0.5]
 [-0.  -0.   0.5]]
```

atztogo:
@jaemolihm, could you tell me what "time reversal" means?

jaemolihm:
I mean spin flip.

atztogo:
Spin flip is restricted to the collinear case (`tensor_rank = 0`). I can implement for the non-colllinear case if you help me to test it. But what is the expected behaviour?

For example, currently following two options are implemented,

with an on-site vector `m=(m_x, m_y, m_z)` at the atomic point `r` and a space group operation `(R|t)`.

```
m' = R m
```
and
```
m' = det(R) R m
```

where `m'` is the on-site vector at `r' = (R|t) r`.

jaemolihm:
I see. Thank you for the explanation. So spin flip is not currently implemented for non-collinear case. Sorry for my confusion.

I was thinking about symmetries where `m' = -Rm` or `m' = -det(R) R m`.

lan496:
Now PR https://github.com/spglib/spglib/pull/162 , which will solve this issue, has merged to `develop`.
[...]

PR #162, opened 2022-06-11
Return time reversal operations for scalar and vectors

lan496:
This is the second PR for supporting magnetic space groups (MSG) in spglib.

- In the python interface, `get_symmetry` with magmoms ≠ None, now returns the `time_reversals` array. True for time-reversal operations, and False for ordinary operations.
- The functions in `spin.c` are fixed and refactored. Now Type-II MSG, which has same-spatial-part operations like g1 and g1’, can be handled.
- Adding an implementation note for MSG detection (`doc/magnetic_spacegroup.rst`). But, I am not sure where to put it.
- Currently, `symprec` is used to compare magmoms. It will be more flexible to add a tolerance parameter for them like magmom_tolerance.

Related: https://github.com/spglib/spglib/issues/150

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,6 +6,7 @@ repos:
   - id: clang-format
     args: ["-i", "--style=file"]  # Use style set in .clang-format
     files: ^src/
+    exclude: src/msg_database.c  # Skip too large file to wait every time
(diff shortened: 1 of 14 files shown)


atztogo:
@lan496, thanks!

PR #168, opened 2022-08-01
Run C-library test and pytest on CI 

lan496:
- Add setting to run Github actions when PR to `develop` branch is opened (and updated)
- Add workflow to run CMake test with Address Sanitizer


atztogo:
Merged. Thanks @lan496!

PR #173, opened 2022-08-04
Reorder arguments of `spg_get_magnetic_spacegroup_type_from_symmetry`

lan496:
Change argument position of `num_operations` to the right behind
time_reversals, because it is more intuitive.


atztogo:
Thanks @lan496!

PR #176, opened 2022-08-06
Implement spg_get_spacegroup_type_from_symmetry

atztogo:
`spg_get_spacegroup_type_from_symmetry` is implemented to replace `spg_get_hall_number_from_symmetry` that is deprecated at v2.0.

- [x] C-API
- [x] test.c
- [x] Python-API
- [x] pytest

### DIFF
--- a/python/test/test_spacegrouop_type_from_symmetry.py
+++ b/python/test/test_spacegrouop_type_from_symmetry.py
@@ -1,0 +1,61 @@
+"""Test of spacegroup_type_from_symmetry."""
+import os
+from pathlib import Path
+from typing import Callable, List
+
+from spglib import (
+    get_spacegroup_type,
+    get_spacegroup_type_from_symmetry,
+    get_symmetry_dataset,
+)
+
+data_dir = os.path.dirname(os.path.abspath(__file__))
+
+
+def test_spacegroup_type_from_symmetry(all_filenames: List[Path], read_vasp: Callable):
+    """Test spacegroup_type_from_symmetry."""
+    for fname in all_filenames:
+        cell = read_vasp(fname)
+        if "distorted" in str(fname):
+            dataset = get_symmetry_dataset(cell, symprec=1e-1)
+            spgtype = get_spacegroup_type_from_symmetry(
+                dataset["rotations"],
+                dataset["translations"],
+                lattice=cell[0],
+                symprec=1e-1,
+            )
+            if spgtype["number"] != dataset["number"]:
+                print("%d != %d in %s" % (spgtype["number"], dataset["number"], fname))
+                ref_cell = (
+                    dataset["std_lattice"],
+                    dataset["std_positions"],
+                    dataset["std_types"],
+                )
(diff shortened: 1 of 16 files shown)


src/magnetic_spacegroup.c
```diff
@@ -581,8 +582,8 @@ static Symmetry *get_space_group_with_magnetic_symmetry(
 
     /* (a, b, c) = (a_prim, b_prim, c_prim) @ tmat */
     prim_sym = prm_get_primitive_symmetry(tmat, sym, symprec);
-
-    *spacegroup = spa_search_spacegroup_with_symmetry(prim_sym, symprec);
+    *spacegroup =
+        spa_search_spacegroup_with_symmetry(prim_sym, unit_lat, symprec);
```
atztogo, line 586: Ping to @lan496, I have modified `spa_search_spacegroup_with_symmetry` to accept basis vectors. Originally, unit basis vectors are set inside `spa_search_spacegroup_with_symmetry`. This change affect to the call here.

atztogo:
@lan496
- Added `conftest.py`, please have a look at `test_spacegrouop_type_from_symmetry.py`.
- `python/_spglib.c` was formatted and many lines were modified, but real change is only addition of `py_get_spacegroup_type_from_symmetry`.
- Docstrings of `spglib.py` were updated, and that for `get_magnetic_symmetry_dataset` was slightly modified to use a similar manner (numpy like).

python/test/test_spacegrouop_type_from_symmetry.py
```diff
@@ -0,0 +1,70 @@
+"""Test of spacegroup_type_from_symmetry."""
+import os
+from pathlib import Path
+from typing import Callable, List
```
lan496, line 4: `typing.List` is deprecated since python 3.9 (https://docs.python.org/3/library/typing.html#typing.List).
Enabling future behavior by
```python
```
and replacing `List[Path]` to `list[Path]` may reduce future chores.

c.f. https://peps.python.org/pep-0563/#enabling-the-future-behavior-in-python-3-7

lan496 (commented):
LGTM! I left a tiny comment.

Yesterday, I learned [unittest2pytest](https://github.com/pytest-dev/unittest2pytest), which seems to help migrate bulit-in unittest to pytest.

python/test/test_spacegrouop_type_from_symmetry.py
atztogo: It's useful. Thanks for your information. I will update using it for the tests.

By the way, I used typehint in `spglib.py` of this PR unintentionally, but I should keep `spglib.py` in old python style for a while.

atztogo:
> Yesterday, I learned [unittest2pytest](https://github.com/pytest-dev/unittest2pytest), which seems to help migrate bulit-in unittest to pytest.

For me, manual migration is easy because I wrote the current tests. But I agree, in general, it is a good idea to use those convenient tools.

PR #177, opened 2022-08-08
Remove hard coded version string in `conf.py`

lan496:
Read version from package metadata of python-spglib via
`importlib.metadata`, which requires Python>=3.8. The config for github
actions are also modified to pip-install spglib at building documents.

c.f. https://docs.python.org/3/library/importlib.metadata.html


atztogo (approved):
I think perfect.

atztogo:
> Python>=3.8.

OK!

Merged. Thanks @lan496!

PR #180, opened 2022-08-09
Install spglib pkgconfig into LIBDIR

dirkmueller:
The pkgconfig file is architecture specific, so it needs
to be installed into libdir/pkgconfig. On Red Hat/SUSE
like distributions libdir is actually $prefix/lib64 on
64bit platforms.

Signed-off-by: Dirk Müller <dirk@dmllr.de>


atztogo:
Merged. Thanks @dirkmueller!

Issue #186, opened 2022-08-31
`get_magnetic_symmetry` causes segmentation fault with high symprec


lan496:
The following example causes a segmentation fault in the last line (the structure is adapted from https://github.com/materialsproject/pymatgen/issues/2641).
It seems ordinary symprec (1e-5) is fine, but high symprec (1e-2) causes the segmentation fault from `get_magnetic_symmetry`.

```python
from spglib import get_symmetry, get_magnetic_symmetry

lattice = [
    [3.36017962, 0.0, 0.95244787],
    [1.70747655, 3.84457132, 0.37960516],
    [-0.0117898, -0.29501452, 6.4450749],
]
positions = [
    [0.0, 0.0, 0.0],
    [0.26737473, 0.2232512, 0.24199933],
    [0.73262527, 0.7767488, 0.75800067],
]
numbers = [0, 1, 1]
magmoms = [
    0.005, 0.002, 0.002,
]

cell = (lattice, positions, numbers)
cell_with_magmoms = (lattice, positions, numbers, magmoms)

def check(symprec, angle_tolerance):
    symmetry = get_symmetry(cell, symprec=symprec, angle_tolerance=angle_tolerance)
    print(f"# of symmetry operations: {len(symmetry['rotations'])}")

    magnetic_symmetry = get_magnetic_symmetry(cell_with_magmoms, symprec=symprec, angle_tolerance=angle_tolerance)
    print(f"# of magnetic symmetry operations: {len(magnetic_symmetry['rotations'])}")


print("With moderate symprec")
check(symprec=1e-5, angle_tolerance=-1)

print("With rough symprec")
check(symprec=1e-2, angle_tolerance=-1)
```

PR #187, opened 2022-08-31
Tighten magnetic tolerance for judging type-II MSG

lan496:
Fixes: https://github.com/spglib/spglib/issues/186

When magnetic moments `m` and `-m` are equal up to `mag_symprec`,
m - (-m) = 2m < mag_symprec should be satisfied. Thus, we need to judge
`m` as zero up to `0.5 * mag_symprec`.

### DIFF
--- a/src/magnetic_spacegroup.c
+++ b/src/magnetic_spacegroup.c
@@ -496,10 +496,6 @@ err:
         representatives = NULL;
     }
 
-    if (ref_sg != NULL) {
-        free(ref_sg);
-        ref_sg = NULL;
-    }
     return 0;
 }
 /* Get family space group (FSG) and return its order. */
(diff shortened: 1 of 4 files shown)


lan496:
@atztogo If no problem, I will merge this PR.
I am not sure how small a patch version should be, but this patch should be included in v.2.0.1.

atztogo:
Please go ahead. About non-major version releases, we can release whenever we want. For the release we need to write a sentence at https://github.com/spglib/spglib/blob/develop/doc/releases.md for package maintainers.

lan496:
@atztogo Thanks for the clarification. The title of this PR will be enough for the release note.
I will notify the corresponding pymatgen's issue when the next non-major version releases.

atztogo:
Thanks!

PR #190, opened 2022-09-22
Add issue templates

lan496:


lan496:
@atztogo I have added issue templates for bug reports and feature requests.
Sections in these templates are almost the same as default ones.
If it seems fine for now, I will merge it to develop. Then, these templates will be used when opening new issues.

atztogo:
Thanks @lan496. Please go ahead.

Issue #194, opened 2022-10-28
Segmentation fault in `v2.0.1` for a specific cell input


munrojm:
## Describe the bug

Segmentation fault when calling `spglib.get_symmetry` for a specific cell input after pip installing `v2.0.1`. Previous versions are fine.

## To Reproduce

The script below causes a segmentation fault on Linux and MacOS systems (**works fine on Windows**) with the following GNU debugger output:
```
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00007fffab38f91e in get_symmetry_with_site_tensors (equivalent_atoms=equivalent_atoms@entry=0x555558c8cd40, permutations=permutations@entry=0x7fffffffbc88, primitive_lattice=primitive_lattice@entry=0x555558cf8770, cell=cell@entry=0x555558cdad90, with_time_reversal=with_time_reversal@entry=1, is_axial=is_axial@entry=0, symprec=symprec@entry=0.10000000000000001, angle_tolerance=angle_tolerance@entry=5, mag_symprec=mag_symprec@entry=-1, max_size=2592) at src/spglib.c:1787
```

```python
import spglib #v2.0.1
import numpy as np

# Cell data taken from task "mp-806965" in the Materials Project database.

data = (
    (
        np.array(
            [
                [5.2419100e00, 0.0000000e00, 0.0000000e00],
                [-3.4590000e-03, 5.6005340e00, 0.0000000e00],
                [-2.6184020e00, -1.8789800e00, 1.1148141e01],
            ]
        ),
        np.array(
            [
                [0.730776, 0.31474, 0.510679],
                [0.943931, 0.414234, 0.832868],
                [0.056367, 0.524886, 0.155721],
                [0.829407, 0.834674, 0.662821],
                [0.665078, 0.665214, 0.316142],
                [0.49794, 0.497249, 0.995771],
                [0.332764, 0.334746, 0.669724],
                [0.001174, 0.996383, 0.002809],
                [0.181127, 0.169218, 0.349259],
                [0.697256, 0.577902, 0.159052],
                [0.308683, 0.416793, 0.832813],
                [0.366841, 0.258942, 0.50616],
                [0.325228, 0.466901, 0.290137],
                [0.188427, 0.691028, 0.042737],
                [0.452284, 0.963256, 0.296948],
                [0.318269, 0.200207, 0.036324],
                [0.647716, 0.725678, 0.498078],
                [0.025986, 0.916766, 0.83257],
[...]

PR #195, opened 2022-10-30
Skip to determine time-reversal part for non-overlapped operations

lan496:
Fixes: https://github.com/spglib/spglib/issues/194

Time-reversal parts are searched after determining linear and
translation parts of symmetry operations by `get_dataset`. It
is impossible in theory but we rarely fail to overlap atoms
possibly due to too high symprec. In that case, skip the
symmetry operation.

@atztogo In some edge cases with high `symprec`, a symmetry operation in `DataContainer->exact_structure->Symmetry` does not follow `symprec` (on the other hand, the not-refined operation follows `symprec`).
In the current implementation for MSG search, permutations of sites are computed from `DataContainer->exact_structure->symmetry`. Then, the segmentation fault in https://github.com/spglib/spglib/issues/194 was caused.
This PR just skips such a suspicious operation and does not change space-group search codes.
So, I will merge this if there are no concerns.

Not addressed in this PR, but I think adding permutation information in `DataContainer` will be better (e.g. ``int *DataContainer->exact_structure->permutations``).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -24,9 +24,9 @@ If spglib returns different results by changing `symprec`, `angle_tolerance`, or
 <!-- A clear and concise description of what you expected to happen. -->
 
 ## Environment
- 
-- Python version: 
-- Spglib's version: 
+
+- Python version:
+- Spglib's version:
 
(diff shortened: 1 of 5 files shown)


atztogo:
Thanks @lan496. Looks good to me.

PR #288, opened 2023-06-10
Initial refactor for layer group implementation

lan496:
This PR addresses the following as a by-product of my code reading for the layer group implementation.

- delete unused functions
- add docstrings and comments
- refactor `match_hall_symbol_db`
- add tests for delaynay reduction and layer-group symmetry

### DIFF
--- a/src/delaunay.c
+++ b/src/delaunay.c
@@ -72,12 +72,6 @@ int del_delaunay_reduce(double min_lattice[3][3], const double lattice[3][3],
     return delaunay_reduce(min_lattice, lattice, -1, symprec);
 }
 
-int del_delaunay_reduce_2D(double min_lattice[3][3], const double lattice[3][3],
-                           const int unique_axis, const double symprec) {
-    debug_print("del_delaunay_reduce_2D:\n");
-    return delaunay_reduce_2D(min_lattice, lattice, unique_axis, -1, symprec);
-}
-
 int del_layer_delaunay_reduce_2D(double min_lattice[3][3],
                                  const double lattice[3][3],
                                  const int unique_axis,
@@ -174,6 +168,9 @@ err:
     return 0;
 }
 
+// @brief Search the shortest vectors from the Delaunay set. If aperiodic axis
+//        exists, find the shortest two vectors inside period plane and the
+//        shortest one outside.
 static void get_delaunay_shortest_vectors(double basis[4][3],
                                           const int lattice_rank,
                                           const double symprec) {
@@ -223,11 +220,10 @@ static void get_delaunay_shortest_vectors(double basis[4][3],
                 }
             }
         }
-        /* find the shortest 2 vectors inside period plane and the shortest one
-         * outside */
     } else {
-        for (i = 0; i < 2; i++) {
-            for (j = 0; j < 2; j++) {
+        // Sort among {b1, b2, b1 + b}
+        for (i = 0; i <= 1; i++) {
+            for (j = 0; j <= 1; j++) {
                 if (mat_norm_squared_d3(b[j]) >
                     mat_norm_squared_d3(b[j + 1]) + ZERO_PREC) {
                     mat_copy_vector_d3(tmpvec, b[j]);
@@ -236,8 +232,13 @@ static void get_delaunay_shortest_vectors(double basis[4][3],
                 }
             }
         }
-        for (i = 3; i < 6; i++) {
-            for (j = 3; j < 6; j++) {
+        // Sort among {b3, b4, b2+b3, b3+b1}
+        // TODO: Is this part required?
+        //       - b4 > b3 because b4 = -(b1+b2+b3) and dot(d3, b1+b2) = 0
+        //       - b2+b3 > b3 because dot(b3, b2) = 0
+        //       - b3+b1 > b3 because dot(b3, b1) = 0
+        for (i = 3; i <= 5; i++) {
+            for (j = 3; j <= 5; j++) {
                 if (mat_norm_squared_d3(b[j]) >
                     mat_norm_squared_d3(b[j + 1]) + ZERO_PREC) {
                     mat_copy_vector_d3(tmpvec, b[j]);
@@ -248,7 +249,7 @@ static void get_delaunay_shortest_vectors(double basis[4][3],
         }
     }
 
-    /* for layer, i = 2, |tmpmat| = 0 is automatically dropped */
+    // For layer, |tmpmat| != 0 for i=2 never happen
     for (i = 2; i < 7; i++) {
         for (j = 0; j < 3; j++) {
             tmpmat[j][0] = b[0][j];
@@ -261,11 +262,13 @@ static void get_delaunay_shortest_vectors(double basis[4][3],
                 basis[1][j] = b[1][j];
                 basis[2][j] = b[i][j];
             }
-            break;
+            return;
         }
     }
 }
 
+// If `lattice_rank==2`, assume `basis[2]` (b3) is along aperiodic axis.
+// TODO: Change `lattice_rank` to a more informative variable
 static int delaunay_reduce_basis(double basis[4][3], const int lattice_rank,
                                  const double symprec) {
     int i, j, k, l;
@@ -291,6 +294,7 @@ static int delaunay_reduce_basis(double basis[4][3], const int lattice_rank,
                     }
                     return 0;
                 } else {
+                    // Here, lattice_rank==2 and dot(b3, b4) > 0
                     /* For layer, if b3 lies in a sphere with center -(b1+b2)/2
                      * and radius |b1+b2|/2, */
                     /* dot(b3, b4) <= 0 will be impossible. This should not
@@ -310,7 +314,9 @@ static int delaunay_reduce_basis(double basis[4][3], const int lattice_rank,
     return 1;
 }
 
-/* aperiodic_axis (if exists) is temporarily moved to b3 */
+// @brief Get extended basis for three-dimensional Delaunay reduction.
+//        If exists, `aperiodic_axis` is temporarily moved to b3 (`basis[2]`).
+// @return lattice_rank Rank of `lattice`
 static int get_extended_basis(double basis[4][3], const double lattice[3][3],
                               const int aperiodic_axis) {
     int i, j, lattice_rank;
@@ -345,9 +351,12 @@ static int get_extended_basis(double basis[4][3], const double lattice[3][3],
     return lattice_rank;
 }
 
-/* For Monocli/Oblique, unique axis is aperiodic axis, j and k are periodic */
-/* For Monocli/Rectang, k is aperiodic axis, unique axis and j are periodic */
-/* j and k are delaunay reduced, which can be incomplete for Monocli/Rectang */
+/* For Monoclinic/Oblique, unique axis is aperiodic axis, j and k are periodic
+ */
+/* For Monoclinic/Rectangular, k is aperiodic axis, unique axis and j are
+ * periodic */
+/* j and k are delaunay reduced, which can be incomplete for
+ * Monoclinic/Rectangular */
 static int delaunay_reduce_2D(double red_lattice[3][3],
                               const double lattice[3][3], const int unique_axis,
                               const int aperiodic_axis, const double symprec) {
--- a/src/delaunay.c
+++ b/src/delaunay.c
@@ -52,9 +52,6 @@ static void get_delaunay_shortest_vectors(double basis[4][3],
                                           const double symprec);
 static int get_extended_basis(double basis[4][3], const double lattice[3][3],
                               const int aperiodic_axis);
-static int delaunay_reduce_2D(double red_lattice[3][3],
-                              const double lattice[3][3], const int unique_axis,
-                              const int aperiodic_axis, const double symprec);
 static int delaunay_reduce_basis_2D(double basis[3][3], const int lattice_rank,
                                     const double symprec);
 static void get_delaunay_shortest_vectors_2D(double basis[3][3],
@@ -70,16 +67,6 @@ int del_delaunay_reduce(double min_lattice[3][3], const double lattice[3][3],
     debug_print("del_delaunay_reduce (tolerance = %f):\n", symprec);
 
     return delaunay_reduce(min_lattice, lattice, -1, symprec);
-}
-
-int del_layer_delaunay_reduce_2D(double min_lattice[3][3],
-                                 const double lattice[3][3],
-                                 const int unique_axis,
-                                 const int aperiodic_axis,
-                                 const double symprec) {
-    debug_print("del_layer_delaunay_reduce_2D:\n");
-    return delaunay_reduce_2D(min_lattice, lattice, unique_axis, aperiodic_axis,
-                              symprec);
 }
 
 /* Return 0 if failed */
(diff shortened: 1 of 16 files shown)


lan496:
So, we need to merge https://github.com/spglib/spglib/pull/278 first to fix CI's clang-format version...

atztogo (approved):
@lan496, great. Thanks!

src/delaunay.c
```diff
@@ -345,17 +338,31 @@ static int get_extended_basis(double basis[4][3], const double lattice[3][3],
...
+// @param[in] unique_axis
+//            Two-fold axis or mirror-plane-perpendicular axis
+// @param[in] aperiodic_axis
+// @param[in] symprec
+// @note For Monoclinic/oblique, the unique axis is also the aperiodic axis.
+//       Axes are {j, k, unique_axis(=aperiodic_axis)}.
+//       For Monoclinic/rectangular, axes are {unique_axis, j,
+//       k(=aperiodic_axis)}. j and k are delaunay reduced, which can be
+//       incomplete for Monoclinic/Rectangular
+int del_layer_delaunay_reduce_2D(double red_lattice[3][3],
```
LecrisUT, line 341: If you don't plan at exposing or accessing the function via `extern`, make the function `static`. I.e. if you only use it within the same file

test/README.md
```diff
@@ -40,6 +40,29 @@ TEST(<your_test_suite_name>, <test_name>) {
...
+```json
+{
+    "configurations": [
+        {
+            ...
+            "includePath": [
+                "${workspaceFolder}/**",
+                "${workspaceFolder}/src",
+                "${workspaceFolder}/test",
+                "/opt/homebrew/include/"
```
LecrisUT, line 58: On CLion it is picked up even if it comes from `FetchContent`. It only needs the appropriate `#include` and then it can navigate the code. Is vscode really that flaky?

Anyway add `cmake-build-*/_deps` to use any `FetchContent` dependencies.

test/test_delaunay.cpp
```diff
@@ -0,0 +1,30 @@
+#include <gtest/gtest.h>
+
+extern "C" {
+#include <math.h>
```
LecrisUT, line 4: Don't use `extern "C"` on std libraries. Let the cpp compiler optimize itself.

LecrisUT, line 4: But a side-note, why not use cpp libraries for cpp code?

test/test_symmetry.cpp
```diff
@@ -0,0 +1,39 @@
...
+    cel_set_layer_cell(cell, lattice, positions, types, aperiodic_axis);
+
+    // Bravais group of the two-dimensional lattice is 4/mmm
+    symmetry = sym_get_operation(cell, symprec, angle_tolerance);
+    ASSERT_EQ(symmetry->size, 16);
+
+    sym_free_symmetry(symmetry);
+    cel_free_cell(cell);
+    symmetry = NULL;
+    cell = NULL;
```
LecrisUT, line 40: `NULL` -> `nullptr`. Also you mean to assert it not assign it here.

test/test_delaunay.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include "debug.h"
+#include "delaunay.h"
+}
+
+TEST(test_delaunay, test_delaunay_reduce_layer) {
+    double min_lattice[3][3];
```
LecrisUT, line 11: I'm always paranoid with C code. Fo you need to explicitly deallocate these? I prefer to use `std::vector` to be safe and pass c_pointer to C functions and use reinterpret_cast when needed

LecrisUT (changes requested):

src/delaunay.c
atztogo: @LecrisUT, this function is used in `spacegroup.c`, so I think that without making it `static` is fine. But is there any other point? I ask because I don't see what you mean about 

> If you don't plan at exposing or accessing the function via extern

LecrisUT: Git was reporting that this function was a new one, but after some search, it's just moved. But in that case:
- move the documentation to header file
- synchronize the variable name if possible between source and header
- remove const int while you're editing that section/file

About
> If you don't plan at exposing or accessing the function via extern

I mean anything that is not in a header file should me marked static

test/test_delaunay.cpp
atztogo: Do you comment in general? Or some specific point?

test/test_symmetry.cpp
atztogo: At least for me,  I am not familiar with C++. If we want to write in C, is there any good way to do it?

test/README.md
atztogo: @lan496, I'm using VSCode, too. With this setting, what will happen?

test/test_delaunay.cpp
LecrisUT: Both. In this case with fixed sized arrays like that, you don't need to use de-allocator?

And in general I am paranoid. E.g. if there is thread_local/static pointer and if for example someone uses the python api and the pointer is replaced without proper de-allocation, then that's a sure way to get memory leaks. If we impose and fix the issues specified by the compiler warnings in `-Wall`, `clang-tidy` inspection, and use the necessary sanitizations, then I would be less paranoid about such hypothetical issues.

test/test_symmetry.cpp
LecrisUT: I would say that writing in C++, especially for these tests it is more advantageous because:
- you get a clear trace of how/what fails
- you decouple potential C memory issues, e.g. not pre-allocating pointers from tests result errors

In these case I only have 2 simple suggestions:
- Use [`std::array`](https://en.cppreference.com/w/cpp/container/array) and continue to use the C interface as you would.
  The reason for this is that you allow in the future to use a wider range of built-in tools like range-for loops similar to python, that make the code much easier to read and detangle.
- Check with the IDE's suggestion.

The `NULL` -> `nullptr` is suggested by CLion and that is because the `NULL` definition in C is different then C++'s `nullptr` and C++ uses the latter for enforcing that a pointer is truely null and not a pointer assigned to `(void*)0`. Think of the issues of comparing addresses of empty strings (`&"" == &""`). That is not guaranteed to give `true` and depends on the compiler's optimization.

Other than that there is a genuine error here in that it is supposed to be `ASSERT_EQ(symmetry == nullptr)` and not an assignment `symmetry = nullptr`.

test/test_delaunay.cpp
atztogo: I see your point. It is uneasy to take a balance of feature developments and fixing holes. I want to merge this PR without going far from the original aim of PR.

src/delaunay.c
lan496: I'll move the document to the header file.

test/README.md
lan496: Without `c_cpp_properties.json` setting, I see the following message:
image

lan496: When I add `${workspaceFolder}/cmake-build-release-llvm/_deps/google-test-src/googletest/include"` to `includePath`, the above error message disappears.

test/test_symmetry.cpp
lan496: I see your point. But I want to confine this PR about the algorithmic side for the layer-group implementation for now. After we adopt C11 in develop branch, the corresponding refactoring would be awesome.

test/README.md
LecrisUT: Doesn't VSCode have system/user/project level settings. It should be documented that `/opt/homebrew/include/` be set for the user space, while `${workspaceFolder}/cmake-build*/_deps` are in the project workspace.

But I find it weird that VSCode or a cmake extension does not use `compile_commands.json`. That is the standard way that cmake has been developing in order for the IDEs to use, and that's part of how CLion does not need special configurations.

test/test_delaunay.cpp
LecrisUT: I understand, but at the very least new code should not introduce new holes. Whenever you are touching a part of a code that your IDE/clang-tidy and any other tools tells you to fix, then do a quick fix for that, otherwise the holes are never addressed.

See for example the warnings already displayed by github in the `Files changed` tab. (Amazed that github is able to do that :open_mouth:)

src/pointgroup.c
```diff
@@ -382,19 +383,15 @@ static int is_exist_axis(const int axis_vec[3], const int axis_index);
 static void sort_axes(int axes[3]);
 static void layer_check_and_sort_axes(int axes[3], const int aperiodic_axis);
 
-/* Return pointgroup.number = 0 if failed */
+// @brief Return pointgroup.number = 0 if failed.
+// @param[out] transform_mat
+// @param[in] rotations
+// @param[in] num_rotations
+// @param[in] aperiodic_axis Use `aperiodic_axis=-1` for space group.
 Pointgroup ptg_get_transformation_matrix(int transform_mat[3][3],
```
LecrisUT, line 386: Header documentation or static function here also?

test/test_symmetry.cpp
LecrisUT, line 40:
```suggestion
    ASSERT_EQ(symmetry, nullptr);
    ASSERT_EQ(cell, nullptr);
```

test/test_symmetry.cpp
```diff
@@ -0,0 +1,39 @@
...
+        {0, 0, 1},
+    };
+    double positions[1][3] = {{0, 0, 0}};
+    const int types[1] = {0};
+    const int aperiodic_axis = 2;
+    const double symprec = 1e-5;
+    const double angle_tolerance = -1;
+
+    cell = cel_alloc_cell(size, NOSPIN);
+    cel_set_layer_cell(cell, lattice, positions, types, aperiodic_axis);
```
LecrisUT:
```suggestion
    cell = cel_alloc_cell(size, NOSPIN);
    ASSERT_NE(cell, nullptr);
    ASSERT_EQ(spg_get_error_code(), SPGLIB_SUCCESS);
    cel_set_layer_cell(cell, lattice, positions, types, aperiodic_axis);
```
Has to be manually asserted because you cannot use `EXPECT_NO_THROW(cell = cel_alloc_cell(size, NOSPIN))`

Similar down below as well.

LecrisUT (changes requested):

test/README.md
atztogo: @LecrisUT, do you assume `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON`?

LecrisUT: No, the IDE automatically does it in its own temporary directory under the configured build directory. It also uses [`cmake-file-api`](https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html#manual:cmake-file-api(7)) to complement and bootstrap the process. You can generally see if the IDE is using that in `${build_dir}/.cmake/api/v1/{query,reply}`.

But in general this is completely transparent to the user and they don't need to setup any such custom variables.

test/test_delaunay.cpp
atztogo: Can `invalid use of pointers to arrays with different qualifiers in ISO C before C2X [-Wpedantic]` and `pointers to arrays with different qualifiers are incompatible in ISO C [-Wpedantic]`
be postponed to fix? Or just removing `const` can be a solution?

LecrisUT: I did a bit of experimentation and research and tl;dr: we don't need to address "invalid * with different qualifiers".
There aren't concerning warnings in this PR now and I did a `clang-tidy` test 
```console
$ pre-commit run --hook-stage manual clang-tidy --from-ref origin/develop --to-ref lan496/layer-group-refactor
clang-tidy...............................................................Passed
```
All seem well for this PR at least on this front. There are still the conversations that I have not marked as resolved which should be addressed and I'm all :+1: for now.

<details><summary>Actually concerning warnings</summary>

The actual concerning warnings are not in this
```
src/refinement.c:1451:9: warning: 'mat_multiply_matrix_di3' accessing 72 bytes in a region of size 24
src/spacegroup.c:1150:51: warning: 'j' may be used uninitialized
```

</details>
<details><summary>About "invalid * with different qualifiers"</summary>

According to the wording in the C standard pre C23, it is interpreted that:
```c
const double matrix[2][3]
```
which is a pointer to a (double-typed) pointer has the const such that it is a pointer to a (double-typed) const pointer, instead of the more natural pointer to a (const double-typed) pointer, meaning that in the former you cannot change where a slice is pointing to and in the latter you cannot change the values of the matrix array.

But in reality no compiler is following the logic of the former and instead they all follow the latter which is consistent with C++ standard as well. And in C23 they have changed the wording so that it is correctly interpreted as the latter (values are constant, and not pointers to the array slice is constant). Indeed if you set `CMAKE_C_STANDARD=23` these warnings go away.

Here is a more detailed example of my tests:
```c
#include <stdio.h>

void test_const_matrix(const int matrix[][2]){
    int vector[2] = {2,3};
//    matrix[0][0] = 2;
    matrix = vector;
    matrix = &vector;
}
void test_const_vector(const int vector[2]){
//    vector[0] = 2;
}
void test_matrix(int matrix[][2]){
    int vector[2] = {2,3};
    matrix[0][0] = 2;
    matrix = vector;
    matrix = &vector;
}
void test_vector(int vector[2]){
    vector[0] = 2;
}

int main() {
    int t_matrix[2][2] = {0};
    int t_vector[2] = {0};
    test_const_matrix(t_matrix);
    test_const_vector(t_vector);
    test_matrix(t_matrix);
    test_vector(t_vector);
    printf("Hello, World!\n");
    return 0;
}
```
Here `test_const_matrix` would fail as expected with `matrix[0][0]=2` even though the wording would indicate otherwise even when I bump down the standard. Although I did not test with antiquated `gcc` compilers to confirm this behaviour is true in old compilers as well. (Note that in the example given there are additional warnings about invalid operations that should be considered).

</details>

I'll leave this conversation unresolved for now, just for visibility.

atztogo: Thanks @LecrisUT.

For `warning: 'mat_multiply_matrix_di3' accessing 72 bytes in a region of size 24`, I have really no idea.

For `warning: 'j' may be used uninitialized`, this is OK, which is confirmed by assert.

LecrisUT: > For `warning: 'j' may be used uninitialized`, this is OK, which is confirmed by assert.

Not really, I will explain in a PR  where I address the clang-tidy issues.

test/test_symmetry.cpp
LecrisUT: Oh sorry, I am wrong, free_cell does not change the pointer to `nullptr`. Can revert this change, I'll make a unit-test for these later

test/test_symmetry.cpp
```diff
@@ -26,6 +26,8 @@ TEST(test_symmetry, test_get_lattice_symmetry_layer) {
     const double angle_tolerance = -1;
 
     cell = cel_alloc_cell(size, NOSPIN);
+    ASSERT_NE(cell, nullptr);
+    ASSERT_EQ(spg_get_error_code(), SPGLIB_SUCCESS);
```
LecrisUT, line 30: Keep these ones though :)

atztogo:
Now I can not follow the current status. To merge this PR, what is left to do and what will be postponed?

lan496:
After the all tests pass and @LecrisUT approves, this PR will be ready to be merged.

LecrisUT (approved):

atztogo:
Thanks very much @lan496 and @LecrisUT!

PR #355, opened 2023-11-09
Fix example wrong results in definition.md

singularitti:
The example code given in [docs](https://spglib.readthedocs.io/en/latest/definition.html#crystallographic-choice-and-rigid-rotation) behave differently compared to the results from the latest version of `spglib`. Please check whether the results was wrong or there are bugs in the code.

`Spglib` version: 2.1.0

### DIFF
--- a/doc/definition.md
+++ b/doc/definition.md
@@ -451,7 +451,7 @@ Cartesian coordinates of the given input unit cell.
 The following example of a python script gives a crystal structure of
 Br whose space group type is *Cmce*. The basis vectors
 $(\mathbf{a}, \mathbf{b}, \mathbf{c})$ are fixed by the symmetry
-crystal in the standardization. The C-centrng determines the c-axis,
+crystal in the standardization. The C-centering determines the c-axis,
 and *m* and *c* operations in *Cmce* fix which directions a- and
 b-axes should be with respect to each other axis. This is the first
 one choice appearing in the list of Hall symbols among 6 different
@@ -490,10 +490,10 @@ This python script is saved in the file `example.py`. Then we get
 % python example.py
 Space group type: Cmce (64)
 Transformation matrix:
-    1  0  0
-    0  1  0
-    0  0  1
-Origin shift: 0.000000 0.500000 0.500000
+   1  0  0
+   0  1  0
+   0  0  1
+Origin shift: 0.000000 0.000000 0.000000
 ```
 
 No rotation was introduced in the idealization. Next, we swap a- and c-axes.
@@ -531,9 +531,9 @@ By this,
 % python spglib-example2.py
 Space group type: Cmce (64)
 Transformation matrix:
-    0  0  1
-    0  1  0
-   -1  0  0
+   0  0  1
+   0  1  0
+  -1  0  0
 Origin shift: 0.000000 0.000000 0.000000
 ```
 
@@ -577,10 +577,10 @@ and
 % python spglib-example3.py
 Space group type: Cmce (64)
 Transformation matrix:
-    1  0  0
-    0  1  0
-    0  0  1
-Origin shift: 0.000000 0.000000 0.500000
+   1  0  0
+   0  1  0
+   0  0  1
+Origin shift: 0.000000 0.000000 0.000000
 ```
 
 The transformation matrix is kept unchanged even though the crystal
--- a/doc/definition.md
+++ b/doc/definition.md
@@ -147,8 +147,8 @@ In this example,
 
 $$
 \boldsymbol{P} = \begin{pmatrix}
+\frac{1}{2} & \frac{\bar{1}}{2} & 0 \\
 \frac{1}{2} & \frac{1}{2} & 0 \\
-\frac{\bar{1}}{2} & \frac{1}{2} & 0 \\
 0 & 0 & 1
 \end{pmatrix}.
 $$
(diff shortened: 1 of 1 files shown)


singularitti:
For the "centrng" typo, it seems like you chose the "centring" in your docs, while I wrote "centering", more in line with American English habits. Both are working English words. I could change it to "centring" if you prefer.

LecrisUT:
Please use `centering` as you've done in this PR. If you can find any similar typos that would be great. Also would be great if they were different commits. Thank you very much for catching these.

@atztogo @lan496 I leave it to you to comment on the validity of the contents.

I will look a bit into why the windows builders do not find `asnan`. I think it's a flag needed to specify using static sanitizers

lan496 (approved):
@singularitti Thank you for your careful checks and corrections!

doc/definition.md
```diff
@@ -490,10 +490,10 @@ This python script is saved in the file `example.py`. Then we get
...
 Space group type: Cmce (64)
 Transformation matrix:
-    1  0  0
-    0  1  0
-    0  0  1
-Origin shift: 0.000000 0.500000 0.500000
+   1  0  0
+   0  1  0
+   0  0  1
+Origin shift: 0.000000 0.000000 0.000000
```
lan496, line 496: @atztogo I am curious about why the origin shift has changed. Maybe related to the axis choice for orthogonal space groups based on Euclidean normalizers?

atztogo:
Thanks @singularitti catching typos. I am not good at English, but I tend to use "centring" following the International Table for Crystallography. In literature, I find both, so no problem which is used. How do you think, @lan496?

doc/definition.md
atztogo: I think you are correct. a(0, 0, 0) and b(1/2, 1/2, 1/2) are in the same Wyckoff set. Plus centring, we get (0, 0, 1/2).
https://www.cryst.ehu.es/cgi-bin/cryst/programs/nph-normsets?from=wycksets&gnum=64
In the current implementation of spglib, we can not distinguish them, although it would be more natural to give the origin shift (0, 0, 0).

atztogo: The following comment is unrelated to this PR. I leave it because it came to my mind.
I probably tried to find the shortest (under some condition) origin shift in this function.
https://github.com/spglib/spglib/blob/a0263dd2e22be20b39e1196a27f80c88990a09cf/src/refinement.c#L1415

singularitti:
Hi @atztogo, "centring" and "centering" are both valid English words; the former is British, and the latter is American. I can change it to "centring".

LecrisUT:
@singularitti, do you want to handle #359 in this PR? Seems pretty straightforward. Not sure whereto add the tests though, to rtd or to pytes

singularitti:
@LecrisUT, do you mean convert these codes I modified into doc tests? Sure! But it might not be instantaneously done since I've been pretty busy recently.

lan496:
> I tend to use "centring" following the International Table for Crystallography. In literature, I find both, so no problem which is used. How do you think, @lan496?

[The IUCr dictionary](https://dictionary.iucr.org/Center) shows both "centre" and "center". So, I also think both will be fine. Actually, I did not notice ITA tends to use the former. Considering to follow the ITA first, I guess "centring" is better.

I find the current spglib mixes the both: 
https://github.com/search?q=repo%3Aspglib%2Fspglib+centering&type=code
https://github.com/search?q=repo%3Aspglib%2Fspglib%20centring&type=code

LecrisUT:
> @LecrisUT, do you mean convert these codes I modified into doc tests? Sure! But it might not be instantaneously done since I've been pretty busy recently.

I can handle these tomorrow. I want to add spell and link checks while we're at it anyway. You're ok with me merging your commits and adding more on top of that.

> > I tend to use "centring" following the International Table for Crystallography. In literature, I find both, so no problem which is used. How do you think, @lan496?
> 
> [The IUCr dictionary](https://dictionary.iucr.org/Center) shows both "centre" and "center". So, I also think both will be fine. Actually, I did not notice ITA tends to use the former. Considering to follow the ITA first, I guess "centring" is better.
> 
> I find the current spglib mixes the both: https://github.com/search?q=repo%3Aspglib%2Fspglib+centering&type=code https://github.com/search?q=repo%3Aspglib%2Fspglib%20centring&type=code

For me I find the spelling `centring` a bit jarring, particularly since I am more used to the AE pronunciation. The other way around, I don't think it is as problematic

atztogo (approved):

singularitti:
While we are at it, could you please also check [this equation](https://github.com/spglib/spglib/blob/a0263dd2e22be20b39e1196a27f80c88990a09cf/doc/definition.md?plain=1#L148-L154)?

```
\boldsymbol{P} = \begin{pmatrix}
\frac{1}{2} & \frac{1}{2} & 0 \\
\frac{\bar{1}}{2} & \frac{1}{2} & 0 \\
0 & 0 & 1
\end{pmatrix}.
```

image

If I understand the symbols correctly, shouldn't the correct matrix be its transpose? Then you could have $\mathbf{b} = \frac{1}{2}\mathbf{a}_s +\frac{1}{2}\mathbf{b}_s`.

atztogo:
@singularitti, thanks. It is incorrect. It must be transposed as below,
```
\boldsymbol{P} = \begin{pmatrix}
\frac{1}{2} & \frac{\bar{1}}{2} & 0 \\
\frac{1}{2} & \frac{1}{2} & 0 \\
0 & 0 & 1
\end{pmatrix}.
```

singularitti:
Agreed. I can update it by the way.

atztogo:
Thanks @singularitti!

singularitti:
My pleasure to help!

PR #386, opened 2024-01-02
Drop ASE Atoms-style input

### LINKED ISSUE #385: Drop ASE Atoms-like input for Python interface
@atztogo
[As stated in the document](https://spglib.readthedocs.io/en/latest/python-spglib.html#ase-atoms-like-input-is-deprecated), an ASE Atoms-like input is deprecated. I think we can drop it now (even ASE itself does not use this style). This will make it easy to annotate types in Python interface functions.

lan496:
Closes https://github.com/spglib/spglib/issues/385


atztogo (approved):
Thanks @lan496!

PR #391, opened 2024-01-10
Fix quick for autodoc2 render plugin

lan496:
After merging https://github.com/spglib/spglib/pull/387, I have noticed auto summaries are generated in reStructuredText, not compatible with other myst documents. 
https://spglib.readthedocs.io/en/latest/python-interface.html#version

This PR uses myst as autodoc2 render plugin and uses reStructuredText only for docstrings.


LecrisUT (approved):
Weird, I wouldn't have expected it to affect that

PR #392, opened 2024-01-12
Replace linter and formatter with ruff and mypy

### LINKED ISSUE #364: ci: Modernize the CI tools

lan496:
Closes https://github.com/spglib/spglib/issues/364

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,32 +29,12 @@ repos:
       - id: end-of-file-fixer
       - id: check-yaml
       - id: check-added-large-files
-  - repo: https://github.com/pycqa/flake8
-    rev: 5.0.4
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.1.12
     hooks:
-      - id: flake8
-        files: python\/.*
-        args:
-          - "--max-line-length=88"
-          - "--ignore=E203,W503"
-  - repo: https://github.com/psf/black
-    rev: 22.6.0
-    hooks:
-      - id: black
-        files: python\/.*
-        args: [ "--line-length=88" ]
-  - repo: https://github.com/pycqa/pydocstyle
-    rev: 6.1.1
-    hooks:
-      - id: pydocstyle
-        files: python/spglib
-        additional_dependencies: [ "toml" ]
-  - repo: https://github.com/pycqa/isort
-    rev: 5.12.0
-    hooks:
-      - id: isort
-        files: python\/.*
-        args: [ "--profile", "black" ]
+      - id: ruff
+        args: [ "--fix", "--show-fixes" ]
+      - id: ruff-format
   - repo: https://github.com/executablebooks/mdformat
     rev: 0.7.16
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,6 +35,13 @@ repos:
       - id: ruff
         args: [ "--fix", "--show-fixes" ]
       - id: ruff-format
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.8.0
+    hooks:
+    - id: mypy
+      pass_filenames: false
+      additional_dependencies:
+        - types-PyYAML
   - repo: https://github.com/executablebooks/mdformat
     rev: 0.7.16
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,7 +7,7 @@ exclude: src\/msg_database.c
 fail_fast: false
 repos:
   - repo: https://github.com/Takishima/cmake-pre-commit-hooks
-    rev: v1.5.3
+    rev: v1.9.5
     hooks:
       - id: clang-format
         additional_dependencies: [ clang-format >= 16 ]
@@ -23,14 +23,14 @@ repos:
           - '-DSPGLIB_WITH_TESTS=ON'
         stages: [ manual ]
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.3.0
+    rev: v4.5.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-yaml
       - id: check-added-large-files
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.1.12
+    rev: v0.1.13
     hooks:
       - id: ruff
         args: [ "--fix", "--show-fixes" ]
@@ -43,7 +43,7 @@ repos:
       additional_dependencies:
         - types-PyYAML
   - repo: https://github.com/executablebooks/mdformat
-    rev: 0.7.16
+    rev: 0.7.17
     hooks:
       - id: mdformat
         additional_dependencies:
@@ -53,14 +53,14 @@ repos:
     # TODO: Temporarily separate docs formatiing
     # https://github.com/executablebooks/mdformat/issues/234
   - repo: https://github.com/executablebooks/mdformat
-    rev: 0.7.16
+    rev: 0.7.17
     hooks:
       - id: mdformat
         additional_dependencies:
           - mdformat-myst
         files: doc\/(?!README).*\.md$
   - repo: https://github.com/python-jsonschema/check-jsonschema
-    rev: 0.21.0
+    rev: 0.27.3
     hooks:
       - id: check-github-workflows
       - id: check-readthedocs
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -186,3 +186,14 @@ extend-ignore = [
 exclude = [
     "database",
 ]
+
+[tool.mypy]
+mypy_path = ["python"]
+files = ["python", "test/functional/python", "example/python_api"]
+python_version = "3.12"
+warn_unused_configs = true
+show_error_codes = true
+# TODO: Enable strict mode once we clean up type annotations
+# strict = true
+enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
+no_strict_optional = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,3 +122,67 @@ command_line = '-m pytest'
 # TODO: Cannot report relative to project path because we cannot do editable installs yet
 #source = ['python/spglib']
 source = ['spglib']
+
+[tool.ruff]
+line-length = 88
+extend-select = [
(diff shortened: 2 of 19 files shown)


.pre-commit-config.yaml
```diff
@@ -23,40 +23,27 @@ repos:
...
+      - id: ruff-format
+        types_or: [python, pyi]
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.8.0
     hooks:
-      - id: isort
-        files: python\/.*
-        args: [ "--profile", "black" ]
+    - id: mypy
+      exclude: "database"
```
lan496: `database/*.py` is not actively maintained now

pyproject.toml
```diff
@@ -122,3 +122,12 @@ command_line = '-m pytest'
...
 source = ['spglib']
+
+[tool.ruff]
+line-length = 88
+exclude = [
+    "database",
+]
+
+[tool.mypy]
+no_strict_optional = true
```
lan496, line 199: mypy is not so smart to pass
```
d: dict | None
assert(d is not None)
d.keys()
```

pyproject.toml
```diff
@@ -122,3 +122,12 @@ command_line = '-m pytest'
 # TODO: Cannot report relative to project path because we cannot do editable installs yet
 #source = ['python/spglib']
 source = ['spglib']
+
+[tool.ruff]
+line-length = 88
+exclude = [
+    "database",
+]
```
LecrisUT: By default `ruff` does not include any linting. It should include something like:
```toml
[tool.ruff.lint]
extend-select = [
    "F",           # pyflakes
    "W",           # pycodestyle-warnings
    "E",           # pycodestyle-errors
    "C90",         # mccabe
    "I",           # isort
    "N",           # pep8-naming
    "D",           # pydocstyle
    "UP",          # pyupgrade
    "YTT",         # flake8-2020
    "ANN",         # flake8-annotations
    "S",           # flake8-bandit
    "BLE",         # flake8-blind-except
    "FBT",         # flake8-boolean-trap
    "B",           # flake8-bugbear
    "A",           # flake8-builtins
    "COM",         # flake8-comma
    "C4",          # flake8-comprehensions
    "T10",         # flake8-debugger
    "EM",          # flake8-errmsg
    "FA",          # flake8-future-annotations
    "ISC",         # flake8-implicit-str-concat
    "ICN",         # flake8-import-conventions
    "G",           # flake8-logging-format
    "PIE",         # flake8-pie
    "T20",         # flake8-print
    "PYI",         # flake8-pyi
    "PT",          # flake8-pytest-style
    "Q",           # flake8-quotes
    "RSE",         # flake8-raise
    "RET",         # flake8-return
    "SLF",         # flake8-self
    "SIM",         # flake8-simplify
    "TID",         # flake8-tidy-imports
    "TCH",         # flake8-type-checking
    "ARG",         # flake8-unused-arguments
    "PTH",         # flake8-use-pathlib
    "TD",          # flake8-todos
    "ERA",         # eradicate
    "PD",          # pandas-vet
    "PGH",         # pygrep-hooks
    "PL",          # pylint
    "TRY",         # tryceratops
    "FLY",         # flynt
    "NPY",         # NumPy
    "PERF",        # perflint
    "FURB",        # refurb
    "RUF",         # Ruff-specific
]
```

test/functional/python/test_spglib.py
```diff
@@ -2,7 +2,7 @@
 import unittest
 
 import numpy as np
-import yaml
```
lan496, line 5: `pip install types-PyYAML` did not work somehow.

python/spglib/spglib.py
```diff
@@ -40,7 +40,7 @@
 import numpy as np
 
 try:
-    from spglib import _spglib as spg
+    from spglib import _spglib as spg  # type: ignore
```
LecrisUT: Why the ignore? Also should scope ignores

LecrisUT (changes requested):
Also, just make sure when you apply `pre-commit` it's a separate commit, and we don't squash commit for this PR

pyproject.toml
LecrisUT: It did on various projects :thinking:. I don't see anything special in the configure either:
```toml
[tool.mypy]
files = ["src", "test"]
mypy_path = ["$MYPY_CONFIG_FILE_DIR/src"]
python_version = "3.12"
warn_unused_configs = true
show_error_codes = true
enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
strict = true
disallow_untyped_defs = false
disallow_incomplete_defs = false

[[tool.mypy.overrides]]
module = ["fmf_jinja.*"]
disallow_untyped_defs = true
disallow_incomplete_defs = true
```

test/functional/python/test_spglib.py
LecrisUT: It has to be in `pre-commit-config.yaml`

.pre-commit-config.yaml
LecrisUT: Better to move the exclude to `pyproject.toml`

.pre-commit-config.yaml
```diff
@@ -23,40 +23,27 @@ repos:
...
-      - id: pydocstyle
-        files: python/spglib
-        additional_dependencies: [ "toml" ]
-  - repo: https://github.com/pycqa/isort
-    rev: 5.12.0
+      - id: ruff
+        args: [ --fix, --exit-non-zero-on-fix ]
+        types_or: [python, pyi]
+      - id: ruff-format
+        types_or: [python, pyi]
```
LecrisUT:
```suggestion
      - id: ruff
        args: ["--fix", "--show-fixes"]
      - id: ruff-format
```
Don't think `type_or`, and `--exit-non-zero-on-fix` does anything in this case.

.pre-commit-config.yaml
lan496: Without it, `pre-commit run --all` complains about `dataset/ptg2symbol.py`. git-commit and `pre-commit run --all` seem to use different settings.

LecrisUT: Even when you set [`exclude`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-exclude)? I'll have to try that locally

pyproject.toml
lan496: Do you recommend these configs that you write?

test/functional/python/test_spglib.py
lan496: Thanks!

pyproject.toml
LecrisUT: Not very experienced with mypy, but that's what I have on one project, and I think I've taken from `scikit-build-core` and `tmt`. So I am not sure if some of them are redundant, but otherwise, looking at the options set, they make sense to have. Worth trying them out as a starting point anyway

LecrisUT:
Also worth referencing the `scienitific-python` configuration: https://github.com/scientific-python/cookie/blob/main/pyproject.toml

python/spglib/spglib.py
lan496: Without ignoring, mypy says
```
error: Module "spglib" has no attribute "_spglib"  [attr-defined]
```

LecrisUT: Yeah, but that's a genuine issue:
- We should be using relative paths there
- We should be providing a `_spglib.pyi` to type-hint the C interface, even if only internally

pyproject.toml
lan496: Flake8's rules seem to be enabled by default. I've added isort's rules.

.pre-commit-config.yaml
lan496: Yes, `pre-commit run --all` ignores it...

pyproject.toml
LecrisUT: Only `E` iirc. Best to be extensive. Here's a list of rules: https://docs.astral.sh/ruff/rules/

I would encourage to keep all of them, but comment out those that are not ready to be added.

pyproject.toml
lan496: The strict mode is too strict now. But, I'll add other flags because they looks useful.

python/spglib/spglib.py
lan496: The relative path did not persuade mypy
```
    from .spglib import _spglib as spg
# -> Module "spglib.spglib" has no attribute "_spglib"  [attr-defined]
```

pyproject.toml
LecrisUT: I would add `strict` as a comment with a todo/open PR to address it.

lan496: Ok, I will leave todo as comment.

LecrisUT:
Please split the "add mypy/ruff configuration" and "apply mypy/ruff” as separate commits in 31b2842840989df073e3bd35342c7cbdae8b90f2. You can run `git commit --no-verify` to skip the pre-commit

LecrisUT (commented):
:+1: for the current setup as an initial implementation. Mind if I rebase your commits a bit before merging?

.pre-commit-config.yaml
```diff
@@ -39,7 +39,7 @@ repos:
     rev: v1.8.0
     hooks:
     - id: mypy
-      exclude: database
+      pass_filenames: false
```
LecrisUT: The size of the python bindings is very small, so might as well run mypy on all sources defined in `pyproject.toml`, i.e. running
```console
$ mypy --ignore-missing-imports --scripts-are-modules
```

lan496:
@LecrisUT  No problem. I'll leave the rest to you.

LecrisUT (approved):

LecrisUT:
Sorry, had to change the committer because I've split the commits. Btw, can you add `docs/readthedocs.org`, and `testing-farm:fedora-39` (and later `fedora-38` as well) to the required workflows to pass? Otherwise if we use `auto-merge` now it would keep those checks dangling.

Also make sure to merge this with a merge commit or rebase to keep the history

lan496:
Thanks to tidy commits. I've updated the branch protection rule. Is it OK?
image

LecrisUT:
Yep, that looks good. I'll bother you again later when I enable Fedora38

PR #530, opened 2024-10-02
Expose `spg_get_symmetry_from_database` in the Fortran interface

pcazade:
I added a simple Fortran wrapper for spg_get_symmetry_from_database based on spg_get_symmetry wrapper.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,6 +11,12 @@ GitHub release pages and in the git history.
 
 ## \[Unreleased\]
 
+### Fortran API
+
+- Expose `spg_get_symmetry_from_database`.
+
+Users can now access the rotation and translation operations from the database, provided the Hall number.
+
 ## v2.5.0 (9 Jul. 2024)
 
 ### Main changes
--- a/fortran/spglib_f08.F90
+++ b/fortran/spglib_f08.F90
@@ -23,6 +23,7 @@ module spglib_f08
             & spg_get_stabilized_reciprocal_mesh, &
             & spg_get_error_code, spg_get_error_message, &
             & spg_get_spacegroup_type, &
+            & spg_get_symmetry_from_database, &
             & spg_get_magnetic_spacegroup_type, &
             & version, version_full, commit, &
             & spg_get_version, spg_get_version_full, spg_get_commit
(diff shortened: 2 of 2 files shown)


LecrisUT (approved):
I guess it's fine. It's not ideal that you need to do the transpose manually on the Fortran side, but it's the fastest way around it.

LecrisUT:
One final change if you may, between the lines:
https://github.com/spglib/spglib/blob/d9c572d1986b0eccc4ed5f18c8d827729a320313/ChangeLog.md?plain=1#L12-L14

Can you add a small snippet like:
```markdown
### Fortran API

- Expose `spg_get_symmetry_from_database`.
```

pcazade:
Will do. Probably not today as I am fully booked until 8 pm (9 pm CET).

I will check for the matrix transposition. I did not realize they were
transposed in the dataset. in my current implementation of spglib in CP2K,
I am using spg_get_symmetry, so it seemed the most natural to use as a
comparison. As I intend to make full use of the dataset, the matrix
transposition is a critical issue. Thanks for pointing this out. I will
update spg_get_symmetry_from_database accordingly.

atztogo:
> it seemed the most natural to use as a comparison.

For example `spg_get_symmetry`, too, the data is just passed without rearranging the data order in memory.

pcazade:
Hi Cristian and Atsushi,

Unless the c_f_pointer function from iso_c_binding transposes the matrices,
it seems to me that the rotation and translation matrices in the dataset
are in the same arrangement as in spg_get_symmetry and the new function I
created:

Here from the dataset module
call c_f_pointer(dset_c%rotations, rotations, shape=[3, 3, n_operations])

Here from the function I added
integer(c_int), intent(out) :: rotation(3, 3, *)

The tests I made with both my new function and the dataset module give the
same results for the rotations. In both cases, the rotation matrices are
transposed compared to their C counterparts. I compared them with the
rotation matrices contained in the Python dataset.

The function I added should therefore be fine as it is consistent with
other modules/functions in the Fortran interface.

Regards,
Pierre

LecrisUT:
Indeed, consistency wise it is fine, it's just the usability point it is not ideal because we are putting a burden on the user to do such transposition. This is of course something that should be handled by us more generally. I will finally have more time as my contract ends, and I'll come back to contribute more to the project and provide a more modern object orientated interface.

For this PR, it's only the `Changelog.md` snippet and we can merge it :)

pcazade:
I added the comment in the Changelog.md as you asked.

I agree with you that, ideally, the end users should not have to think about the array arrangement within the library. That said, it is not always bad to have the matrices already transposed, depending on the linear algebra operations done with them.

Thanks for your help.

LecrisUT:
@lan496 can you overwrite the test requirement? I am not sure why F39 failed there, but it will be EOL in a few weeks, so you can just those test requirements.

lan496:
@LecrisUT 
Sorry for being late. I've removed F39 for test requirements
image

PR #592, opened 2025-06-21
PR to fix (python) get_spacegroup_type_from_symmetry

atztogo:
This PR will fix the bug in the python interface of `get_spacegroup_type_from_symmetry`. The bug is that `lattice` is not transposed.


LecrisUT (approved):

atztogo:
Comment: Theoretically it should be unnecessary to rely on basis vectors, but currently it is due to the specific implementation of space group type search in spglib.

PR #593, opened 2025-06-21
Update document for (spg_)get_spacegroup_type_from_symmetry

atztogo:
~~@LecrisUT, I could not build document locally. Could you tell me what is the best way to check the doc build?~~

#584 revealed that the `lattice` parameter of `(spg_)get_spacegroup_type_from_symmetry` is necessary to transform symmetry operations in a reasonable form when the matrix representations of those operations were obtained with respect to oblique basis vectors. Following this, the documentation has been updated.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -378,13 +378,17 @@ SpglibSpacegroupType spg_get_spacegroup_type_from_symmetry(
 );
 ```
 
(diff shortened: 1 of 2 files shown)


atztogo:
Sorry, I found `docs/README.md` and it worked.

lan496 (approved):

PR #612, opened 2025-11-04
Update implementation of PR-610 and add the test

atztogo:
This PR is a simplified version of PR #610. A new test was added to validate the changes of PR #610 since I had forgotten about it. Locally, I verified that this test fails on the commit before #610.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,6 +13,9 @@ GitHub release pages and in the git history.
 
 ## Main changes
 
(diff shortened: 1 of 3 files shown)


atztogo:
@lan496, I think this implementation is simpler and more intuitive than #610. But do you think this implementation is correct?

lan496 (approved):
LGTM for choosing clean basis vectors in getting space group types of XSG and FSG before identifying MSG. I think PR 610 did the latter

PR #626, opened 2025-12-22
Drop building wheel for macos-13

atztogo:
Added windows-11-arm. Can we make it?

### DIFF
--- a/.github/workflows/step_build-wheel.yaml
+++ b/.github/workflows/step_build-wheel.yaml
@@ -26,7 +26,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Build wheels
-        uses: pypa/cibuildwheel@v3.3.0
+        uses: pypa/cibuildwheel@v3.3
         env:
           CIBW_BUILD: ${{ inputs.cibw_build }}
       - uses: actions/upload-artifact@v5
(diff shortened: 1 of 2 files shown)


.github/workflows/step_build-wheel.yaml
```diff
@@ -19,14 +19,14 @@ jobs:
...
-        os: [ ubuntu-latest, ubuntu-24.04-arm, windows-latest, macos-13, macos-latest ]
+        os: [ ubuntu-latest, ubuntu-24.04-arm, windows-latest, windows-11-arm, macos-latest ]
 
     steps:
       - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - name: Build wheels
-        uses: pypa/cibuildwheel@v3.3
+        uses: pypa/cibuildwheel@v3.3.0
```
LecrisUT:
```suggestion
        uses: pypa/cibuildwheel@v3.3
```
Let's not fix the versions so much. This allows us to pick up some bug fixes automatically since the tag is moving as patches are updated.

.github/workflows/step_test.yaml
LecrisUT, line 31: Can you also add the `windows-11-arm` here too?
```yaml
          - os: windows-11-arm
            toolchain: windows
```

LecrisUT (commented):
Cool coverage for the windows-11-arm 👍

LecrisUT (approved):

